PR #1, opened 2022-10-20
create a processor and store for incoming traces

CtrlSpice:
### otlp payload processor:
When an otlp payload comes in, it is wrapped in a ptrace.Traces object where the data is structured roughly as follows:

- resource spans
  - resource 
    - resource-specific data
    - resource attributes
  - instrumentation scope spans
    - scope
      - scope-specific data
      - scope attributes
    - spans
      - span-specific data
      - span attributes

The otlp payload processor extracts each span and attaches the relevant scope and resource to this span in order to create a data structure in line with what we want to display on the client side of the application.

### trace store
This is a persistent data store that consists of a map of spans indexed by their trace IDs, and a queue that limits the number of traces kept in memory (currently hardcoded to 10 000 for convenience, though this should eventually be moved to the config file).


desktop-exporter/trace_store.go
```diff
@@ -0,0 +1,73 @@
...
+	store.mut.Lock()
+	defer store.mut.Unlock()
+
+	// Enqueue, then append, as the enqueue process checks if the traceID is already in the map to keep the trace alive
+	store.enqueueTrace(spanData.TraceID)
+	store.traceMap[spanData.TraceID] = append(store.traceMap[spanData.TraceID], spanData)
+}
+
+func (store *TraceStore) enqueueTrace(traceID string) {
+	// Make room for the trace in the queue if need be
```
jmorrell, line 38:
```suggestion
	// If we have exceeded the maximum number of traces we plan to store
	// make room for the trace in the queue by deleting the oldest trace
```

desktop-exporter/trace_store.go
```diff
@@ -0,0 +1,73 @@
...
+	_, traceIDExists := store.traceMap[traceID]
+	if traceIDExists {
+		e := store.findQueueElement(traceID)
+		if e == nil {
+			panic(errors.New("traceID mismatch between TraceStore.traceMap and TraceStore.traceQueue"))
+		}
+
+		store.traceQueue.MoveToBack(e)
+	} else {
+		//Enqueue traceID
```
jmorrell, line 55:
```suggestion
		// Add traceID to the back of the queue with the most recent traceIDs
```

desktop-exporter/trace_store.go
```diff
@@ -0,0 +1,73 @@
...
+	store.traceMap[spanData.TraceID] = append(store.traceMap[spanData.TraceID], spanData)
+}
+
+func (store *TraceStore) enqueueTrace(traceID string) {
+	// Make room for the trace in the queue if need be
+	for store.traceQueue.Len() >= MAX_QUEUE_SIZE {
+		store.dequeueTrace()
+	}
+
+	// If the traceID is already in the queue, move it to the back of the line
```
jmorrell, line 43: This seems backwards to me, but I'm not sure why. Leaving a comment explaining that this is the case is 💯

jmorrell (changes requested):
I think we should refactor `extractSpans` but otherwise this looks great!

I'd also expect tests for both `extractSpans` and `TraceStore`, but these can be done in follow-on PRs

This is a bit large for one PR as it's adding 2 new things. For the next time you encounter a situation like this: I would suggest splitting this into a PR that add the processing logic + tests first, followed by a PR that adds the datastore + tests. But I think we can move forward with this one as-is.

desktop-exporter/otlp_payload_processor.go
```diff
@@ -0,0 +1,81 @@
...
+	Attributes             map[string]interface{}
+	DroppedAttributesCount uint32
+}
+
+type ScopeData struct {
+	Name                   string
+	Version                string
+	Attributes             map[string]interface{}
+	DroppedAttributesCount uint32
+}
```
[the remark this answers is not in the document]
CtrlSpice: This is not done yet. There is more span data (events, attributes, links, dropped counts, etc...) to bring in, but I wanted to see if I could work out the extraction/storage logic with a partial span before finishing up.

desktop-exporter/trace_store.go
```diff
@@ -0,0 +1,73 @@
...
+	mut        sync.Mutex
+	traceQueue *list.List
+	traceMap   map[string][]SpanData
+}
+
+func NewTraceStore() *TraceStore {
+	return &TraceStore{
+		mut:        sync.Mutex{},
+		traceQueue: list.New(),
+		traceMap:   make(map[string][]SpanData),
```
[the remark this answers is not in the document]
CtrlSpice: Cool! I did not know that was a thing, but it makes sense

desktop-exporter/trace_store.go
CtrlSpice: I will switch it the right way around in the next pr.

desktop-exporter/desktop_exporter.go
```diff
@@ -13,50 +10,34 @@ import (
...
-
-	exporter.accumulator.add(context, traces)
-
-	buf, err := exporter.tracesMarshaler.MarshalTraces(traces)
-	if err != nil {
-		return err
+func (exporter *desktopExporter) pushTraces(ctx context.Context, traces ptrace.Traces) error {
+	spans := extractSpans(ctx, traces)
+	for _, span := range spans {
+		exporter.traceStore.Add(ctx, span)
```
jmorrell, line 20: so much nicer :D

desktop-exporter/trace_store.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// If the traceID is already in the queue, move it to the back of the line
+	// TODO: change the direction of the queue to this section of code more readable
+	_, traceIDExists := store.traceMap[traceID]
+	if traceIDExists {
+		e := store.findQueueElement(traceID)
+		if e == nil {
+			panic(errors.New("traceID mismatch between TraceStore.traceMap and TraceStore.traceQueue"))
+		}
```
jmorrell, line 51: Reading this just now I thought that `e` was short for `error`, not `element`, and then was confused about the `e == nil` check. Renaming it to `element` might improve readability here

desktop-exporter/trace_store.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (store *TraceStore) enqueueTrace(traceID string) {
+	// If we have exceeded the maximum number of traces we plan to store
+	// make room for the trace in the queue by deleting the oldest trace
+	for store.traceQueue.Len() >= MAX_QUEUE_SIZE {
+		store.dequeueTrace()
+	}
+
+	// If the traceID is already in the queue, move it to the back of the line
+	// TODO: change the direction of the queue to this section of code more readable
```
jmorrell, line 45: I tend to recommend not leaving TODO's in code. Inevitably they will be the longest-lived code you ever write. Ask me how I know 😅

jmorrell (approved):
:shipit: 💯

PR #3, opened 2022-10-26
Payload processor test

CtrlSpice:
First, we ensure that all the statically generated data in testdata/trace is extracted correctly by the payload processor, and is reflected in the resulting spans.

Then we loop through the spans and derive the indices assigned to the resources/scopes/spans at creation from the index of the resulting span, thus verifying that each span is attached to the correct resource and scope.


desktop-exporter/testdata/trace.go
```diff
@@ -0,0 +1,81 @@
...
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+var (
+	spanStartTimestamp = pcommon.NewTimestampFromTime(time.Date(2022, 10, 21, 7, 10, 2, 100, time.UTC))
+	spanEventTimestamp = pcommon.NewTimestampFromTime(time.Date(2020, 10, 21, 7, 10, 2, 150, time.UTC))
+	spanEndTimestamp   = pcommon.NewTimestampFromTime(time.Date(2020, 10, 21, 7, 10, 2, 300, time.UTC))
+)
+
+func GenerateTraces(resourceCount, scopeCount, spanCount int) ptrace.Traces {
```
jmorrell, line 16:
```suggestion
func GenerateOTLPPayload(resourceCount, scopeCount, spanCount int) ptrace.Traces {
```

desktop-exporter/otlp_payload_processor_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	resourceCount = 2
+	scopeCount    = 3
+	spanCount     = 4
+)
+
+func TestExtractSpans(t *testing.T) {
+	traces := testdata.GenerateTraces(resourceCount, scopeCount, spanCount)
+	spans := extractSpans(context.Background(), traces)
```
jmorrell, line 22: an assertion against `spans.length` would be a good addition I think

jmorrell (approved):
Great job! :shipit: ❤️

desktop-exporter/otlp_payload_processor_test.go
CtrlSpice: Good call!

PR #4, opened 2022-10-26
Complete SpanData

CtrlSpice:
Here we add fields to the SpanData struct so that it reflects the data found in ptrace.Span.

Things added: links, events, status code and message, span kind, trace state, and a dropped counter for attributes, events, and links.

The oltp_payload_processor_test was then updated to cover these additions.

### DIFF
--- a/desktop-exporter/otlp_payload_processor.go
+++ b/desktop-exporter/otlp_payload_processor.go
@@ -21,15 +21,15 @@ type ScopeData struct {
 }
 
 type SpanData struct {
-	TraceID    string
-	TraceState string
-
-	SpanId       string
+	TraceID      string
+	TraceState   string
+	SpanID       string
 	ParentSpanID string
-	Name         string
-	Kind         string
-	StartTime    time.Time
-	EndTime      time.Time
+
+	Name      string
+	Kind      string
+	StartTime time.Time
+	EndTime   time.Time
 
 	Attributes map[string]interface{}
 	Events     []EventData
@@ -141,7 +141,7 @@ func aggregateSpanData(span ptrace.Span, eventData []EventData, LinkData []LinkD
 		TraceID:    span.TraceID().HexString(),
 		TraceState: span.TraceState().AsRaw(),
 
-		SpanId:       span.SpanID().HexString(),
+		SpanID:       span.SpanID().HexString(),
 		ParentSpanID: span.ParentSpanID().HexString(),
 		Name:         span.Name(),
 		Kind:         span.Kind().String(),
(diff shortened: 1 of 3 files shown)


desktop-exporter/otlp_payload_processor.go
```diff
@@ -8,26 +8,54 @@ import (
...
+	Links      []LinkData
+	Resource   *ResourceData
+	Scope      *ScopeData
+
+	DroppedAttributesCount uint32
+	DroppedEventsCount     uint32
+	DroppedLinksCount      uint32
+
+	StatusCode    string
+	StatusMessage string
```
jmorrell, line 45: This is just to say that the way you broke up or ordered your attributes here makes intuitive sense to me and is pleasing and nice to read, but I could not tell you why

desktop-exporter/otlp_payload_processor.go
```diff
@@ -8,26 +8,54 @@ import (
...
+	Attributes             map[string]interface{}
+	DroppedAttributesCount uint32
+}
+
 type SpanData struct {
-	TraceID      string
+	TraceID    string
+	TraceState string
+
 	SpanId       string
```
jmorrell, line 27: nit: `SpanId` should be written like `TraceID`, like `SpanID`.

I think that the go linter will call things like this out, so we should hook that up soon.

desktop-exporter/otlp_payload_processor.go
```diff
@@ -45,14 +73,34 @@ func extractSpans(_ context.Context, traces ptrace.Traces) []SpanData {
...
+	return eventDataSlice
+}
+
+func extractLinks(links ptrace.SpanLinkSlice) []LinkData {
+	linkDataSlice := make([]LinkData, 0, links.Len())
+	for linkIndex := 0; linkIndex < links.Len(); linkIndex++ {
+		linkDataSlice = append(linkDataSlice, aggregateLinkData(links.At(linkIndex)))
+	}
+
+	return linkDataSlice
```
jmorrell, line 101: Nope, I can't think of a cleaner way to write this. This is the way

image

desktop-exporter/otlp_payload_processor.go
jmorrell: nit: Maybe `SpanID` and `ParentSpanID` should be grouped with `TraceID` and `TraceState` though

jmorrell (approved):
Such a nice clean PR! A pleasure to review ☺️

beautiful-amazing

desktop-exporter/otlp_payload_processor.go
CtrlSpice: Oh dear. Yes it definitely should and I have no idea how I haven't noticed it yet. Good call on the linter.

PR #6, opened 2022-10-29
make the maximum trace store queue length configurable via argument

CtrlSpice:
There are 2 reasons for this change:

1. Being able to test the trace store with a manageably short queue
2. In preparation for making the trace queue configurable via the config.yaml file in a future update


jmorrell (approved):
🚀

PR #8, opened 2022-10-31
remove references to the logging tool 

CtrlSpice:
These were artefacts from the log exporter I used as a basis for the structure of this project, and are no longer in use. Deleting them lets me have a clean slate for my server config.


jmorrell (approved):
Red Diffs are Best Diffs ❤️

PR #9, opened 2022-11-01
Implement a basic go server with access to the traceStore and routing functionality

CtrlSpice:
At this stage, the server logs the contents of the traceStore when it receives a request by the "/traces" path, and logs the spans associated with a specific traceID when it receives a request by "/trace/{id}"


desktop-exporter/server.go
```diff
@@ -0,0 +1,70 @@
...
+
+type Server struct {
+	server     http.Server
+	traceStore *TraceStore
+}
+
+func getTracesHandler(traceStore *TraceStore) func(http.ResponseWriter, *http.Request) {
+	return func(writer http.ResponseWriter, request *http.Request) {
+		traceStore.mut.Lock()
+		defer traceStore.mut.Unlock()
```
jmorrell, line 19: In a future PR we should define some methods to read data and push this logic into the store itself, though this is perfectly fine for the moment

desktop-exporter/server.go
```diff
@@ -0,0 +1,70 @@
...
+
+func getTraceIDHandler(traceStore *TraceStore) func(http.ResponseWriter, *http.Request) {
+	return func(writer http.ResponseWriter, request *http.Request) {
+		traceStore.mut.Lock()
+		defer traceStore.mut.Unlock()
+
+		writer.WriteHeader(http.StatusOK)
+
+		traceID := mux.Vars(request)["id"]
+		spans, err := json.Marshal(traceStore.traceMap[traceID])
```
jmorrell, line 40: There is an error case here where the trace id doesn't exist. Again, fine for now while we're sketching stuff out, but something we should be aware of when this handler grows up

desktop-exporter/server.go
```diff
@@ -0,0 +1,70 @@
...
+		if err != nil {
+			fmt.Printf("error marshalling trace %s: %s\n", traceID, err)
+		} else {
+			fmt.Fprintf(writer, "Hello trace %s:\n%s\n", traceID, spans)
+		}
+	}
+}
+
+func NewServer(traceStore *TraceStore) *Server {
```
jmorrell, line 49: I'm not 100% sure why this needs to be a `*Server` instead of a `Server` but if it works I am happy to not ask too many questions

desktop-exporter/factory.go
```diff
@@ -38,6 +38,5 @@ func createTracesExporter(ctx context.Context, set component.ExporterCreateSetti
 		exporterhelper.WithRetry(exporterhelper.RetrySettings{Enabled: false}),
 		exporterhelper.WithQueue(exporterhelper.QueueSettings{Enabled: false}),
 		exporterhelper.WithStart(desktopExporter.Start),
-		exporterhelper.WithShutdown(desktopExporter.Shutdown),
```
jmorrell, line 41: I think we will want this back, but we can revisit shutting things down cleanly later.

jmorrell (approved):
For PRs like this where there are no tests, I would suggest pasting a terminal session of spinning up and using this server so the review can "see" it working, either the raw text or a screencast

PR #10, opened 2022-11-02
serve a static page with an alert

CtrlSpice:
Add a static directory from which we can serve content and make sure everything works.
[static-web.pdf](https://github.com/CtrlSpice/otel-desktop-viewer/files/9924406/static-web.pdf)


jmorrell (approved):
🕸️

PR #11, opened 2022-11-03
Set up makefile and js build

CtrlSpice:
- Added a makefile with some useful commands, including one to watch and build our javascript app.
- Added main.jsx and main.css for the front-end portion of the app
- Updated gitignore to exclude build artefacts from the repo


jmorrell (approved):
Can we build it? Yes we can! 🧑‍🏭

PR #12, opened 2022-11-03
Serve json representation of traceMap and log it to console on the client side

CtrlSpice:
- In server, set header Content-Type to "application/json" to make sure the client expects json in response
- Update from ReactDOM.render to root.render, as the latter is no longer supported as of [React v18.](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis)
- Log the contents of traceMap to the console

json-response

### DIFF
--- a/desktop-exporter/app/main.jsx
+++ b/desktop-exporter/app/main.jsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import ReactDOM from 'react-dom'
+import {createRoot} from 'react-dom/client'
 
 function App() {
   return (
@@ -9,4 +9,9 @@ function App() {
   );
 }
 
-ReactDOM.render(<App />, document.querySelector('#root'));
+// ReactDOM.render is no longer supported in React 18 - using createRoot instead
+const container = document.getElementById('root')
+const root = createRoot(container)
+root.render(<App />)
+
+fetch("/traces").then(response => response.json()).then(data => console.log(data))
--- a/desktop-exporter/app/main.jsx
+++ b/desktop-exporter/app/main.jsx
@@ -9,7 +9,6 @@ function App() {
   );
 }
 
-// ReactDOM.render is no longer supported in React 18 - using createRoot instead
 const container = document.getElementById('root')
 const root = createRoot(container)
 root.render(<App />)
--- a/desktop-exporter/server.go
+++ b/desktop-exporter/server.go
@@ -22,7 +22,7 @@ func getTracesHandler(traceStore *TraceStore) func(http.ResponseWriter, *http.Re
 		if err != nil {
 			panic(fmt.Errorf("error marshalling traceStore: %s\n", err))
 		} else {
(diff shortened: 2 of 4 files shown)


desktop-exporter/app/main.jsx
```diff
@@ -9,4 +9,9 @@ function App() {
   );
 }
 
-ReactDOM.render(<App />, document.querySelector('#root'));
+// ReactDOM.render is no longer supported in React 18 - using createRoot instead
```
jmorrell, line 12: This is good context to have for the PR, but I don't think this comment should live in the code forever.

desktop-exporter/app/main.jsx
```diff
@@ -9,4 +9,9 @@ function App() {
...
+const container = document.getElementById('root')
+const root = createRoot(container)
+root.render(<App />)
+
+fetch("/traces").then(response => response.json()).then(data => console.log(data))
```
jmorrell, line 17: We've piped the data all the way to the browser! 🥳

desktop-exporter/server.go
```diff
@@ -18,13 +18,13 @@ func getTracesHandler(traceStore *TraceStore) func(http.ResponseWriter, *http.Re
 		traceStore.mut.Lock()
 		defer traceStore.mut.Unlock()
 
-		writer.WriteHeader(http.StatusOK)
-
-		traces, err := json.Marshal(traceStore.traceMap)
+		jsonTraces, err := json.Marshal(traceStore.traceMap)
 		if err != nil {
-			fmt.Printf("error marshalling traceStore: %s\n", err)
+			panic(fmt.Errorf("error marshalling traceStore: %s\n", err))
```
jmorrell, line 23: This probably doesn't merit a `panic` but we can iterate on our endpoints later.

desktop-exporter/server.go
```diff
@@ -18,13 +18,13 @@ func getTracesHandler(traceStore *TraceStore) func(http.ResponseWriter, *http.Re
...
 		} else {
-			fmt.Fprintf(writer, "Hello traceStore:\n%s\n", traces)
+			writer.WriteHeader(http.StatusCreated)
```
jmorrell, line 25:
```suggestion
			writer.WriteHeader(http.StatusOK)
```

`http.StatusCreated` is usually used as a response code to a `POST` request creating / storing some bit of data. Here we are fetching data with a `GET` request, so a `200` or `http.StatusOK` is what we want.

https://go.dev/src/net/http/status.go

jmorrell (approved):

desktop-exporter/app/main.jsx
CtrlSpice: Is gone now!

desktop-exporter/server.go
CtrlSpice: Ok!

PR #13, opened 2022-11-05
Add functions to retrieve data from the TraceStore

CtrlSpice:
- GetLatestTraceIDs retrieves the TraceIDs of the most recent spans added to the TraceStore. If the number of IDs requested exceeds the number of traces in the store, GetLatestTraceIDs will return the entire contents of the store.
- GetTraceByID retrieves a slice of spans associated with a TraceID. If the TraceID is not found, the function returns an error
- Test functions have been added to trace_store_test to cover both new functions.

### DIFF
--- a/desktop-exporter/trace_store.go
+++ b/desktop-exporter/trace_store.go
@@ -32,6 +32,35 @@ func (store *TraceStore) Add(_ context.Context, spanData SpanData) {
 	store.traceMap[spanData.TraceID] = append(store.traceMap[spanData.TraceID], spanData)
 }
 
+func (store *TraceStore) GetRecentTraceIDs(traceCount int) []string {
+	store.mut.Lock()
+	defer store.mut.Unlock()
+
+	if traceCount > store.traceQueue.Len() {
+		traceCount = store.traceQueue.Len()
+	}
+
+	recentTraceIDs := make([]string, 0, traceCount)
+	element := store.traceQueue.Front()
+
+	for i := 0; i < traceCount; i++ {
+		recentTraceIDs = append(recentTraceIDs, element.Value.(string))
+		element = element.Next()
+	}
+
+	return recentTraceIDs
+}
+
+func (store *TraceStore) GetTraceByID(traceID string) ([]SpanData, error) {
+	trace, traceExists := store.traceMap[traceID]
+
+	if !traceExists {
+		return nil, errors.New("traceID not found")
+	}
+
+	return trace, nil
+}
+
 func (store *TraceStore) enqueueTrace(traceID string) {
 	// If the traceID is already in the queue, move it to the front of the line
 	_, traceIDExists := store.traceMap[traceID]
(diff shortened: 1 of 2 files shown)


desktop-exporter/trace_store_test.go
```diff
@@ -89,3 +89,63 @@ func TestAddExceedingTraceLimits(t *testing.T) {
...
+	recentTraceIDs := store.GetRecentTraceIDs(numRecentTraceIDs)
+
+	// Validate that the number of IDs returned is equal to the lesser value of:
+	// - The number of IDs requested or
+	// - The number traces available in the store
+	if totalTraces < numRecentTraceIDs {
+		assert.Len(t, recentTraceIDs, totalTraces)
+	} else {
+		assert.Len(t, recentTraceIDs, numRecentTraceIDs)
+	}
```
jmorrell, line 118: Why is this a conditional when these values are fixed constants? I would expect a test for each case

desktop-exporter/trace_store_test.go
```diff
@@ -89,3 +89,63 @@ func TestAddExceedingTraceLimits(t *testing.T) {
...
+	// Assign each span a TraceID derived from its index before adding it to the store
+	// This TraceID is passed as an argument to test *TraceStore.GetTrace
+	for i, span := range spans {
+		span.TraceID = strconv.Itoa(i)
+		store.Add(ctx, span)
+	}
+
+	// Verify that we are able to retrieve every trace in the store by its TraceID
+	for i := 0; i < totalTraces; i++ {
+		trace, _ := store.GetTraceByID(strconv.Itoa(i))
```
jmorrell, line 144:
```suggestion
		trace, err := store.GetTraceByID(strconv.Itoa(i))
		assert.NoError(t, err)
```

jmorrell (approved):
🚀

desktop-exporter/trace_store_test.go
CtrlSpice: Why does github hot have a facepalm emoji? This is the assert I needed, yes.

desktop-exporter/trace_store.go
```diff
@@ -32,6 +32,35 @@ func (store *TraceStore) Add(_ context.Context, spanData SpanData) {
...
+
+	for i := 0; i < traceCount; i++ {
+		recentTraceIDs = append(recentTraceIDs, element.Value.(string))
+		element = element.Next()
+	}
+
+	return recentTraceIDs
+}
+
+func (store *TraceStore) GetTraceByID(traceID string) ([]SpanData, error) {
```
jmorrell, line 54: It might be good to create an output type here that we can add more data to rather than an array. Something like:

```go
type struct TraceData {
  Spans []SpanData,
  DurationMS int64,
  RootServiceName string,
  // extend here
}

func (store *TraceStore) GetTraceByID(traceID string) (TraceData, error) {
```

or maybe this should live in the server? 🤔 What do you think?

PR #14, opened 2022-11-09
Refactor serializable types

CtrlSpice:
Moved serializable types into their own file and added json annotations.

### DIFF
--- a/desktop-exporter/types.go
+++ b/desktop-exporter/types.go
@@ -15,7 +15,7 @@ type TraceSummary struct {
 
 type TraceData struct {
 	Spans      []SpanData `json:"spans"`
-	DurationMS int64      `json:"duration"`
+	DurationMS int64      `json:"durationMS"`
 }
 
 type ResourceData struct {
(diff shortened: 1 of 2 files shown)


desktop-exporter/types.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type TraceSummary struct {
+	TraceID   string `json:"traceID"`
+	SpanCount uint32 `json:"spanCount"`
+}
+
+type TraceData struct {
+	Spans      []SpanData `json:"spans"`
+	DurationMS int64      `json:"duration"`
```
jmorrell, line 18: In general I find it very useful to include the unit in the name for durations when you don't have a type system to convey that information

```suggestion
	DurationMS int64      `json:"durationMS"`
```

Alternatively you could send along the start and end timestamps and calculate the duration on the client 🤔

desktop-exporter/types.go
```diff
@@ -0,0 +1,71 @@
+package desktopexporter
```
jmorrell, line 1: Something about having types laid out all organized like this makes me very happy

jmorrell (approved):

PR #17, opened 2022-11-12
Update TraceStore.traceMap to store *TraceData

CtrlSpice:
Update `TraceStore.traceMap` so that the spans associated with a traceID key are wrapped in a `TraceData` struct. This makes it both more convenient to marshal as json, and easier to extend in the future. Tests updated accordingly.


jmorrell (approved):
Factoring this out into its own data type should make the next steps a lot cleaner 💯

PR #20, opened 2022-11-14
Update tracesHandler 

CtrlSpice:
- Renamed the function `tracesHandler` for simplicity
- Now using the request's query string to determine how many recent traces we want to show. If filter is not set, currently shows everything. This is for my own development convenience for now, and we may wish to set a reasonable default in the future.
- `tracesHandler` builds summaries for the recent traces we want to display, stores them in `RecentSummaries.TraceSummaries`, marshals the `RecentSummaries` wrapper, and sends it off.

 
traces

traces?result=3

### DIFF
--- a/desktop-exporter/server.go
+++ b/desktop-exporter/server.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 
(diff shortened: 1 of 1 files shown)


jmorrell (approved):
It's got screenshots and everything 😍 

3a27f370797171 5baf4dd46cfe9

PR #21, opened 2022-11-14
Update traceIDHandler to serve TraceData

CtrlSpice:
So, this works (see screenshot), but I would like some advice on making the error handling a bit more elegant. Not really sure how to do that?

traces:{id}

### DIFF
--- a/desktop-exporter/server.go
+++ b/desktop-exporter/server.go
@@ -51,19 +51,22 @@ func tracesHandler(store *TraceStore) func(http.ResponseWriter, *http.Request) {
 	}
 }
 
-func getTraceIDHandler(traceStore *TraceStore) func(http.ResponseWriter, *http.Request) {
+func traceIDHandler(store *TraceStore) func(http.ResponseWriter, *http.Request) {
 	return func(writer http.ResponseWriter, request *http.Request) {
-		traceStore.mut.Lock()
-		defer traceStore.mut.Unlock()
-
 		traceID := mux.Vars(request)["id"]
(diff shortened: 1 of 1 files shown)


desktop-exporter/server.go
```diff
@@ -51,27 +51,30 @@ func tracesHandler(store *TraceStore) func(http.ResponseWriter, *http.Request) {
...
 	return func(writer http.ResponseWriter, request *http.Request) {
-		traceStore.mut.Lock()
-		defer traceStore.mut.Unlock()
-
 		traceID := mux.Vars(request)["id"]
-		jsonTrace, err := json.Marshal(traceStore.traceMap[traceID])
+		traceData, err := store.GetTrace(traceID)
 		if err != nil {
 			fmt.Println(err)
+			writer.WriteHeader(http.StatusInternalServerError)
```
jmorrell, line 60: Something like this maybe?

```go
traceData, err := store.GetTrace(traceID)
if err != nil {
  fmt.Println(err)
  writer.WriteHeader(http.StatusInternalServerError)
  return
}

jsonTraceData, err := json.Marshal(traceData)
if err != nil {
  fmt.Println(err)
  writer.WriteHeader(http.StatusInternalServerError)
  return
}

writer.WriteHeader(http.StatusOK)
writer.Header().Set("Content-Type", "application/json")
writer.Write(jsonTraceData)
```

CtrlSpice: Thank you this is exactly what I was looking for! 

axolotl

jmorrell (approved):
:shipit: 

icegif-279

PR #23, opened 2022-11-17
create basic trace summary loader

CtrlSpice:
Create a loader that requests `TraceSummaries` from the server and renders the `traceID`s for each trace as links.

### DIFF
--- a/desktop-exporter/app/routes/main-view.jsx
+++ b/desktop-exporter/app/routes/main-view.jsx
@@ -1,20 +1,31 @@
 import React from 'react';
-import { Link } from "react-router-dom";
+import { Link, useLoaderData } from "react-router-dom";
+
+export async function loader() {
+    const response = await fetch("/traces");
+    const traceSummaries = await response.json();
+    return traceSummaries;
+}
 
 export default function MainView() {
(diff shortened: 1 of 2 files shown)


desktop-exporter/app/routes/main-view.jsx
```diff
@@ -1,20 +1,31 @@
 import React from 'react';
-import { Link } from "react-router-dom";
+import { Link, useLoaderData } from "react-router-dom";
+
+export async function loader() {
```
jmorrell, line 4:
```suggestion
export async function mainLoader() {
```

If you change the name here, the import in `main.jsx` will look like:

```js
import MainView, { mainLoader } from './routes/main-view';
```

which feels a little more sensible to me.

jmorrell: or you could rename everything else to be consistent with the `summaryLoader` and it could be:

```js
import SummaryView, { summaryLoader } from './routes/summary-view';
```

jmorrell (approved):
DALL·E 2022-11-16 16 44 44 - an axolotl riding a cargo bike

PR #25, opened 2022-11-17
Fetch and log individual traceData

CtrlSpice:
- When you click on a trace id in the `MainView` summary table, the `traceLoader` retrieves the `TraceData` associated with this id from the server.
- In this iteration, we display the traceID from the first span and log the rest, like so:

traces:{traceID}

### DIFF
--- a/desktop-exporter/app/routes/trace-view.jsx
+++ b/desktop-exporter/app/routes/trace-view.jsx
@@ -1,13 +1,20 @@
 import React from 'react';
-import { useParams, Link } from 'react-router-dom';
+import { useParams, Link, useLoaderData } from 'react-router-dom';
+
+export async function traceLoader({ params }) {
+    const response = await fetch(`/traces/${params.traceID}`);
+    const traceData = await response.json();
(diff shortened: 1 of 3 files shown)


desktop-exporter/app/routes/trace-view.jsx
```diff
@@ -1,13 +1,20 @@
...
 
 export default function TraceView() {
-    let { traceID } = useParams();
+    const { spans } = useLoaderData();
+    console.log(spans);
     return (
         <>
             <Link to={"/"}>Back</Link>
-            <p>I am Item {traceID}</p>
+            <p>I am Item {spans[0].traceID}</p>
```
jmorrell, line 17: We should probably add `TraceID` to `TraceData` soon https://github.com/CtrlSpice/otel-desktop-viewer/blob/main/desktop-exporter/types.go#L22

jmorrell (approved):
DALL·E 2022-11-17 17 04 10 - chibi corgi giving a thumbs up while standing on a laptop

PR #26, opened 2022-11-19
Render span data in a table

CtrlSpice:
Using `react-window`

traces:{traceID}

### DIFF
--- a/desktop-exporter/app/routes/trace-view.jsx
+++ b/desktop-exporter/app/routes/trace-view.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Link, useLoaderData } from 'react-router-dom';
-import { FixedSizeList as List } from 'react-window';
+import { FixedSizeList } from 'react-window';
 
 
 export async function traceLoader({ params }) {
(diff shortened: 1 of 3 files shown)


desktop-exporter/app/routes/trace-view.jsx
```diff
@@ -1,20 +1,34 @@
 import React from 'react';
-import { useParams, Link, useLoaderData } from 'react-router-dom';
+import { Link, useLoaderData } from 'react-router-dom';
+import { FixedSizeList as List } from 'react-window';
```
jmorrell, line 3: Opinion: I usually prefer to leave imports as their original name unless there is a conflict, so I would leave this as `FixedSizeList`, and I would rename it to something like `ReactWindowList` for clarity if there was a conflict rather than to something more generic.

desktop-exporter/package.json
```diff
@@ -9,5 +9,8 @@
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-router-dom": "^6.4.3"
+    },
+    "dependencies": {
+        "react-window": "^1.8.8"
```
jmorrell, line 14: nit: this is only used at build-time, so this should be in `devDependencies`

jmorrell (approved):
DALL·E 2022-11-17 17 01 27 - chibi corgi

PR #27, opened 2022-11-19
Add TraceID field to TraceData struct

CtrlSpice:
So we don't need to do `trace.Spans[0].TraceID` to access it anymore.


jmorrell (approved):
DALL·E 2022-11-19 12 58 08 - a tiny orange traffic cone has a big smile while sitting on a stump, digital painting

PR #28, opened 2022-11-19
rename api paths

CtrlSpice:
Renamed api paths to `api/traces` and `api/traces/{traceID}` to avoid conflict with the app's routes

### DIFF
--- a/desktop-exporter/app/routes/main-view.jsx
+++ b/desktop-exporter/app/routes/main-view.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Link, useLoaderData } from "react-router-dom";
 
 export async function mainLoader() {
-    const response = await fetch("/traces");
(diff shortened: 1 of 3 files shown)


jmorrell (approved):
This looks 💯 and should simplify things

We are still likely going to run into problems when reloading, say the trace detail page since `/traces/124` will not serve `index.html`, so the static server will return a `404`.

There are a number of approaches we can use here: https://stackoverflow.com/questions/27928372/react-router-urls-dont-work-when-refreshing-or-writing-manually/36623117#36623117

but given the small # of routes we're managing, the simplest is probably to list all possible routes on the server and serve `index.html`. Currently there are only two: `/traces` and `/traces/:id`.

Or use hash routing for the client side routing instead of `pushState`

DALL·E 2022-11-19 13 17 33 - a corgi, Johannes Vermeer

PR #31, opened 2022-11-23
Add server-side handler for /traces/{id}

CtrlSpice:
...that serves index.html and lets the client populate the data. This means refreshing the page doesn't break it anymore 🥳


jmorrell (approved):
DALL·E 2022-11-23 16 10 40 - chibi corgi giving a thumbs up, watercolor and pencil

PR #33, opened 2022-11-26
Add prettier and format all the things

jmorrell:
This brings in [prettier](https://prettier.io/) as a dependency, adds a handy `format-js` target to the Makefile, and applies the prettier config to all the existing code. The diff is large, but all the changes are programmatic except for those in `Makefile` and `desktop-exporter/package.json`.

### DIFF
--- a/desktop-exporter/package.json
+++ b/desktop-exporter/package.json
@@ -20,7 +20,7 @@
     },
     "prettier": {
         "jsxSingleQuote": false,
-        "printWidth": 100,
+        "printWidth": 80,
         "semi": true,
         "singleAttributePerLine": true,
         "singleQuote": false,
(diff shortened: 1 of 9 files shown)


desktop-exporter/package.json
```diff
@@ -5,16 +5,26 @@
...
+        "usehooks-ts": "^2.9.1"
+    },
+    "prettier": {
+        "jsxSingleQuote": false,
+        "printWidth": 80,
+        "semi": true,
+        "singleAttributePerLine": true,
+        "singleQuote": false,
+        "tabWidth": 2,
+        "trailingComma": "all"
```
jmorrell, line 28: Formatting options live here. If there any changes or [any more you want to add](https://prettier.io/docs/en/options.html) please let me know :)

CtrlSpice: Ooooh! This is good. I like the defaults for all the rest, so no need to add anything.

CtrlSpice (approved):
DALL·E 2022-11-17 10 36 58

PR #34, opened 2022-11-26
Add typescript

jmorrell:
Depends on #33 

- Add typescript + some type definitions as `devDependencies`
- Change the file extensions to `.tsx` from `.jsx` 
- Update makefile so that we can build and run type checks

This does not validate type checking yet, just gets the tooling in place.

```
❯ make validate-typescript
cd desktop-exporter; npx tsc --noEmit
app/components/header.tsx:12:24 - error TS7006: Parameter 'props' implicitly has an 'any' type.

12 export function Header(props) {
                          ~~~~~

app/error-page.tsx:13:13 - error TS18046: 'error' is of type 'unknown'.

13         <i>{error.statusText || error.message}</i>
               ~~~~~

app/error-page.tsx:13:33 - error TS18046: 'error' is of type 'unknown'.

13         <i>{error.statusText || error.message}</i>
                                   ~~~~~

app/main.tsx:27:25 - error TS2345: Argument of type 'HTMLElement | null' is not assignable to parameter of type 'Element | DocumentFragment'.
  Type 'null' is not assignable to type 'Element | DocumentFragment'.

27 const root = createRoot(container);
                           ~~~~~~~~~

app/routes/main-view.tsx:12:16 - error TS7031: Binding element 'index' implicitly has an 'any' type.

12 function Row({ index, style, data }) {
                  ~~~~~

app/routes/main-view.tsx:12:23 - error TS7031: Binding element 'style' implicitly has an 'any' type.

12 function Row({ index, style, data }) {
                         ~~~~~

app/routes/main-view.tsx:12:30 - error TS7031: Binding element 'data' implicitly has an 'any' type.

12 function Row({ index, style, data }) {
                                ~~~~

app/routes/main-view.tsx:23:18 - error TS7006: Parameter 'props' implicitly has an 'any' type.

23 function Sidebar(props) {
                    ~~~~~

app/routes/main-view.tsx:37:11 - error TS2339: Property 'traceSummaries' does not exist on type 'unknown'.

37   const { traceSummaries } = useLoaderData();
             ~~~~~~~~~~~~~~

app/routes/trace-view.tsx:7:37 - error TS7031: Binding element 'params' implicitly has an 'any' type.

7 export async function traceLoader({ params }) {
                                      ~~~~~~

app/routes/trace-view.tsx:13:25 - error TS7031: Binding element 'index' implicitly has an 'any' type.

13 function WaterfallRow({ index, style, data }) {
                           ~~~~~

[...]

PR #35, opened 2022-11-26
Add typescript annotations to make typechecking pass

jmorrell:
Depends on #34 

- Adds annotations for the API responses in `desktop-exporter/app/types/api-types.ts`
- Adds annotations for each react component props
- Makes type checking pass

```
otel-desktop-viewer git/add-typescript-annotations
❯ make validate-typescript
cd desktop-exporter; npx tsc --noEmit

```

### DIFF
--- a/desktop-exporter/app/routes/trace-view.tsx
+++ b/desktop-exporter/app/routes/trace-view.tsx
@@ -13,7 +13,7 @@ export async function traceLoader({ params }: any) {
 
 type WaterfallRowProps = {
   index: number;
-  style: {};
+  style: React.CSSProperties;
   data: WaterfallViewProps;
 };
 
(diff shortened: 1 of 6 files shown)


desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -64,10 +81,10 @@ End: ${span.endTime}
 }
 
 export default function TraceView() {
-  const traceData = useLoaderData();
-  const [selectedSpanID, setSelectedSpanID] = React.useState(
-    traceData.spans[0].spanID,
-  );
+  const traceData = useLoaderData() as TraceData;
```
jmorrell, line 84: I was not able to get types to flow from the loader function, through the router, through `useLoaderData()`, but we can assert the type here and get typing through the rest of the app.

desktop-exporter/app/types/api-types.ts
```diff
@@ -0,0 +1,66 @@
+export type TraceSummary = {
```
jmorrell, line 1: This is a direct translation of the types in https://github.com/CtrlSpice/otel-desktop-viewer/blob/main/desktop-exporter/types.go

CtrlSpice (approved):
DALL·E 2022-11-28 09 08 55 - chibi corgi looking through a telescope

PR #40, opened 2022-11-29
refactor main view into its component parts

CtrlSpice:
- Refactor `<Sidebar />` and `<TraceList />` into their own files
- Use a hook to set the height of `<FixedSizeList /}` relative to its parent

### DIFF
--- a/desktop-exporter/app/components/sidebar.tsx
+++ b/desktop-exporter/app/components/sidebar.tsx
@@ -1,0 +1,44 @@
+import React from "react";
+import { ArrowLeftIcon, ArrowRightIcon } from "@chakra-ui/icons";
+import { Flex, IconButton, useColorModeValue } from "@chakra-ui/react";
+
+import { TraceList } from "../components/traceList";
+
+type SidebarProps = {
+  isFullWidth: boolean;
+  toggle: () => void;
+};
+
+export function Sidebar(props: SidebarProps) {
+  const sidebarColour = useColorModeValue("pink.100", "pink.900");
+
+  let sidebarWidth = "70px";
+  let buttonIcon = <ArrowRightIcon />;
+  let traceList = <></>;
+
+  if (props.isFullWidth) {
+    sidebarWidth = "250px";
+    buttonIcon = <ArrowLeftIcon />;
+    traceList = <TraceList />;
+  }
+
+  return (
+    <Flex
+      bgColor={sidebarColour}
+      direction="column"
+      transition="width 0.2s ease-in-out"
+      width={sidebarWidth}
+    >
+      <Flex justifyContent="flex-end">
+        <IconButton
+          aria-label="Expand Sidebar"
+          colorScheme="pink"
+          icon={buttonIcon}
+          margin="15px"
+          onClick={props.toggle}
+        />
+      </Flex>
+      {traceList}
+    </Flex>
+  );
+}
--- a/desktop-exporter/app/components/sidebar.tsx
+++ b/desktop-exporter/app/components/sidebar.tsx
@@ -3,10 +3,12 @@ import { ArrowLeftIcon, ArrowRightIcon } from "@chakra-ui/icons";
 import { Flex, IconButton, useColorModeValue } from "@chakra-ui/react";
 
 import { TraceList } from "../components/traceList";
+import { TraceSummary } from "../types/api-types";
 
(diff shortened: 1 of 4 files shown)


desktop-exporter/app/components/traceList.tsx
```diff
@@ -0,0 +1,48 @@
...
+      to={`traces/${data[index].traceID}`}
+      style={style}
+    >
+      {data[index].traceID}
+    </NavLink>
+  );
+}
+
+export function TraceList() {
+  const { traceSummaries } = useLoaderData() as TraceSummaries;
```
jmorrell, line 27: This does work fine, but I think I would recommend requiring this data higher up in the hierarchy and passing it down to `TraceList` as a `prop`.

Ideally the component directly rendered by the router, `MainView` would handle fetching the data from its `loader` so that `TraceList` is more self-contained and doesn't have dependencies on something several levels above itself.

In this way the coupling between `mainLoader` and `MainView` that we can't property type is constrained to one file / component.

```jsx
export default function MainView() {
  let [isFullWidth, setFullWidth] = useBoolean();
  let { traceSummaries } = useLoaderData() as TraceSummaries;
  
  return (
    <div className="container">
      <Sidebar
        isFullWidth={isFullWidth}
        toggle={setFullWidth.toggle}
        traceSummaries={traceSummaries}
      />
      <Outlet />
    </div>
  );
}
```

jmorrell (approved):
DALL·E 2022-11-28 21 13 15 - cute corgi standing next to a telescope with a starry background, anime

desktop-exporter/app/components/traceList.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const ref = useRef(null);
+  const size = useSize(ref);
+
+  return (
+    <Flex
+      ref={ref}
+      height="100%"
+    >
+      <FixedSizeList
+        className="list"
```
jmorrell, line 40: Do we still need this `className`? There are no rules for it left

desktop-exporter/app/routes/main-view.tsx
```diff
@@ -1,99 +1,27 @@
...
-    </Flex>
-  );
-}
-
 export default function MainView() {
   let [isFullWidth, setFullWidth] = useBoolean();
+  let { traceSummaries } = useLoaderData() as TraceSummaries;
 
   return (
     <div className="container">
```
jmorrell, line 20: You could probably replace this with a [`Flex`](https://chakra-ui.com/docs/components/flex) and delete the `.container` CSS rule

```jsx
<Flex height="100vh">
// ...
</Flex>
```

desktop-exporter/app/routes/main-view.tsx
```diff
@@ -1,99 +1,27 @@
...
-          onClick={props.toggle}
-        />
-      </Flex>
-      {traceList}
-    </Flex>
-  );
-}
-
 export default function MainView() {
   let [isFullWidth, setFullWidth] = useBoolean();
```
jmorrell, line 16: We probably want the the sidebar expanded by default instead of collapsed

```suggestion
  let [isFullWidth, setFullWidth] = useBoolean(true);
```

desktop-exporter/app/components/sidebar.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { TraceSummary } from "../types/api-types";
+
+type SidebarProps = {
+  isFullWidth: boolean;
+  toggle: () => void;
+  traceSummaries: TraceSummary[];
+};
+
+export function Sidebar(props: SidebarProps) {
+  const sidebarColour = useColorModeValue("pink.100", "pink.900");
```
jmorrell, line 15: I appreciate your commitment to resisting American English in your naming

jmorrell (approved):
DALL·E 2022-11-30 13 44 52 - cute corgi standing next to a telescope with a starry background, cartoon

desktop-exporter/app/components/sidebar.tsx
CtrlSpice: Thank you. I do try ever so dutifully.

desktop-exporter/app/components/traceList.tsx
CtrlSpice: We do not

PR #41, opened 2022-12-08
Style tracelist

CtrlSpice:
- Style the list of traces in the sidebar to display their traceID and summaries
- Make the background colour of the `LinkBox` reflect which trace is currently selected.
sidebar-select
sidebar-select-dark

### DIFF
--- a/desktop-exporter/app/components/traceList.tsx
+++ b/desktop-exporter/app/components/traceList.tsx
@@ -13,70 +13,73 @@ import { useSize } from "@chakra-ui/react-use-size";
 
 import { TraceSummary } from "../types/api-types";
 
-type RowProps = {
-  index: number;
-  style: Object;
-  data: TraceSummary[];
-};
+ const listItemHeight = 100;
+ const listItemTopMargin = 10;
 
-function Row({ index, style, data }: RowProps) {
-  let isSelected = useMatch(`traces/${data[index].traceID}`);
+ type RowProps = {
+   index: number;
+   style: Object;
+   data: TraceSummary[];
+ };
 
-  let backgroundColour = isSelected
-    ? useColorModeValue("teal.100", "teal.700")
-    : useColorModeValue("whiteAlpha.700", "whiteAlpha.400");
+ function Row({ index, style, data }: RowProps) {
+   let isSelected = useMatch(`traces/${data[index].traceID}`);
 
-  return (
-    <div style={style}>
-      <LinkBox
-        bgColor={backgroundColour}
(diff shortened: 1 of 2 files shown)


desktop-exporter/app/components/traceList.tsx
```diff
@@ -13,13 +20,38 @@ type RowProps = {
 };
 
 function Row({ index, style, data }: RowProps) {
+  let isSelected = useMatch(`traces/${data[index].traceID}`);
+
+  let backgroundColour = isSelected
+    ? useColorModeValue("teal.100", "teal.700")
+    : useColorModeValue("whiteAlpha.700", "whiteAlpha.400");
```
jmorrell, line 27: 👏 👏 👏

desktop-exporter/app/components/traceList.tsx
```diff
@@ -13,13 +20,38 @@ type RowProps = {
...
+        marginX="10px"
+        marginTop="10px"
+        rounded="md"
+      >
+        <Heading
+          marginY="1"
+          noOfLines={1}
+          size="sm"
+        >
+          <LinkOverlay
```
jmorrell, line 44: Just to make sure, do you have to click on the trace id or can you click anywhere in the box?

desktop-exporter/app/components/traceList.tsx
```diff
@@ -40,11 +72,11 @@ export function TraceList(props: TraceListProps) {
         height={size ? size.height : 0}
         itemData={props.traceSummaries}
         itemCount={props.traceSummaries.length}
-        itemSize={50}
+        itemSize={110}
```
jmorrell, line 75: Since you're using `height` and `marginTop` here to calculate this, it might be a good idea to factor those out into constants in this file. That way changing them in the future is simpler and less fussy.

```js
const listItemHeight = 100;
const listItemTopMargin = 10;
```

jmorrell (approved):
DALL·E 2022-12-08 13 22 18

desktop-exporter/app/components/traceList.tsx
CtrlSpice: You click the LinkBox. The link overlay is the link inside the box that gets 'lifted' (not sure of the right word).

desktop-exporter/app/components/traceList.tsx
CtrlSpice: Good call, since these numbers will change once we iron out the info we want in TraceSummary, and what we want to display in this list.

PR #42, opened 2022-12-10
Update sidebar layout and highlight selected trace summary

CtrlSpice:
- Make sidebar look a little less 2008
sidebar-select
sidebar-select-dark

### DIFF
--- a/desktop-exporter/app/components/traceList.tsx
+++ b/desktop-exporter/app/components/traceList.tsx
@@ -13,73 +13,79 @@ import { useSize } from "@chakra-ui/react-use-size";
 
 import { TraceSummary } from "../types/api-types";
 
- const listItemHeight = 100;
- const listItemTopMargin = 10;
+const listItemHeight = 100;
+const listItemBottomMargin = 2;
 
- type RowProps = {
-   index: number;
-   style: Object;
-   data: TraceSummary[];
- };
+type RowProps = {
+  index: number;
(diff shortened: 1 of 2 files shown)


desktop-exporter/app/components/traceList.tsx
```diff
@@ -13,73 +13,79 @@ import { useSize } from "@chakra-ui/react-use-size";
...
-   let backgroundColour = isSelected
-     ? useColorModeValue("teal.100", "teal.700")
-     : useColorModeValue("whiteAlpha.700", "whiteAlpha.400");
+  return (
+    <div style={style}>
+      <LinkBox
+        bgColor={backgroundColour}
+        height="100px"
+        padding="10px"
+        marginBottom={`${listItemBottomMargin}px`}
```
jmorrell, line 38: I think if it's pixels you can just write the value without the string formatting:
```suggestion
        marginBottom={listItemBottomMargin}
```

jmorrell (approved):
DALL·E 2022-12-10 11 30 07 - A corgi as a spectator in an art gallery, digital art

desktop-exporter/app/components/traceList.tsx
CtrlSpice: Chakra interprets just the value as 0.5 rem, and I need the height of the row to stay predictable for react-window's sake.

PR #44, opened 2022-12-13
normalize status code names …

CtrlSpice:
and omit status message (description) when code is unset or ok, per open telemetry specification outlined here : [https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status](url)


jmorrell (approved):
DALL·E 2022-12-13 14 31 33 - A happy corgi with an axolotl on its head

PR #45, opened 2022-12-13
normalize span kind names

CtrlSpice:
to reflect these specifications:
[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind](url)


jmorrell (approved):
DALL·E 2022-12-13 18 02 43 - A happy corgi with an axolotl on its head

PR #46, opened 2022-12-14
style and render events panel in detail view

CtrlSpice:
events-collapsed
events-expanded


jmorrell (approved):
DALL·E 2022-12-14 11 30 12 - a corgi on top of the squamish chief, digital art

PR #50, opened 2023-01-03
First Pass at Ordering Spans for Display in Waterfall View

CtrlSpice:
Spans are ordered according to their parent/child relationships. Additionally, spans that share a `parentSpanID` are sorted by earliest start time. 

This is being done in `TraceView` rather than `WaterfallView` because that is where we set the `selectedSpanID` when changing routes. It was simpler to pull the whole thing up and pass the sorted array to `WaterfallView`. 

Currently, orphan spans and incomplete trees aren't handled. I will do this thing in my next PR.

Also, the depth is visually represented with padding, which is not very clear. I will find a better solution once I have taken care of the orphans. I'm aware that I sound like a supervillain. This is fine. 

OrderedSpansPass1

### DIFF
--- a/desktop-exporter/app/routes/trace-view.tsx
+++ b/desktop-exporter/app/routes/trace-view.tsx
@@ -3,11 +3,13 @@ import { useLoaderData } from "react-router-dom";
 import { Grid, GridItem } from "@chakra-ui/react";
 import { arrayToTree, TreeItem } from "performant-array-to-tree";
 
+import { TraceData } from "../types/api-types";
+import { SpanWithMetadata } from "../types/metadata-types";
+
 import { Header } from "../components/header";
 import { DetailView } from "../components/detail-view/detail-view";
 import { WaterfallView } from "../components/waterfall-view/waterfall-view";
-import { getNsFromString } from "../utils/duration";
-import { TraceData, SpanData } from "../types/api-types";
+import { getNsFromString, getTraceDurationNs } from "../utils/duration";
 
 export async function traceLoader({ params }: any) {
   let response = await fetch(`/api/traces/${params.traceID}`);
@@ -17,18 +19,20 @@ export async function traceLoader({ params }: any) {
 
 export default function TraceView() {
   let traceData = useLoaderData() as TraceData;
+  let traceDurationNs = getTraceDurationNs(traceData.spans);
+
   let spanTree = arrayToTree(traceData.spans, {
     id: "spanID",
     parentId: "parentSpanID",
   });
   let orderedSpans = orderSpans(spanTree[0], 0, []);
   let [selectedSpanID, setSelectedSpanID] = React.useState<string>(
-    orderedSpans.length ? orderedSpans[0].spanID : "",
+    orderedSpans.length ? orderedSpans[0].span.spanID : "",
   );
 
   // if we get a new trace because the route changed, reset the selected span
   React.useEffect(() => {
-    setSelectedSpanID(orderedSpans[0].spanID);
+    setSelectedSpanID(orderedSpans[0].span.spanID);
   }, [traceData]);
 
   let selectedSpan = traceData.spans.find(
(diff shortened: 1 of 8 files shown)


desktop-exporter/app/types/api-types.ts
```diff
@@ -36,6 +36,8 @@ export type SpanData = {
 
   statusCode: string;
   statusMessage: string;
+
+  depth?: number;
```
jmorrell, line 40: I'm a little hesitant to extend this type with new fields given it corresponds 1:1 with OpenTelemetry's data model 🤔 Our UI information is leaking into our type definitions we use to parse API data.

If we're going to add extra data, can we add it into a specific metadata struct?

```typescript
export type SpanMetaData = {
  depth?: number
  // we will likely find other things to go here
}

export type SpanData = {
  /* ... */
  metadata?: SpanMetaData;
}
```

Alternatively, we can give `orderedSpans` a type like this, and avoid altering `SpanData`:

```typescript
type SpanWithMetadata = {
  span: SpanData;
  metadata: SpanMetaData;
}

let orderedSpans = [] as SpanWithMetadata[];
```

```typescript
orderedSpans.push({
  span: parentSpan,
  metadata: { depth }
});
```

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -59,3 +64,25 @@ export default function TraceView() {
...
+  parentNode: TreeItem,
+  depth: number,
+  orderedSpans: SpanData[],
+) {
+  let parentSpan = parentNode.data;
+  parentSpan.depth = depth;
+  orderedSpans.push(parentSpan);
+
+  let children = parentNode.children.sort(
+    // Not sure if this is how you write this in typescript, but I did my best
```
jmorrell, line 78: I'm not sure either 🤷 but this seems fine

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -59,3 +64,25 @@ export default function TraceView() {
...
+    (a: { data: { startTime: string } }, b: { data: { startTime: string } }) =>
+      getNsFromString(a.data.startTime) - getNsFromString(b.data.startTime),
+  );
+
+  children.forEach((node: TreeItem) => {
+    orderedSpans = orderSpans(node, depth + 1, orderedSpans);
+  });
+
+  return orderedSpans;
+}
```
jmorrell, line 88: I think this might be clearer written without recursion, using a stack, and likely easier to handle the orphans.

This is a quick sketch:

```typescript
for (let root of spanTree) {
  let stack = [{
    node: root.data,
    depth: 0,
  }];

  while (stack.length) {
    let {node, depth} = stack.pop();

    orderedSpans.push({
      span: node,
      depth: depth
    })

    node.children
      .sort(...)
      .forEach(child => stack.push({ node: child.data, depth: depth + 1}));
  }
}
```

jmorrell (changes requested):
Sorted spans! This is great

desktop-exporter/app/types/api-types.ts
CtrlSpice: I went with the second solution, to avoid leaking our UI into our type definitions, especially since I expect to generate more metadata around drawing the tree that connects the spans. It took a bit of refactoring though, so I'm sorry in advance that the changes are spread across a bunch of files. 

I also moved `getTraceDurationNs` into `TraceView`, and passed it down as a prop, because we have a perfectly good array of SpanData right there, rather than separating all the spans from their metadata at `WaterfallView` level just to do this one thing. I'm not sure if this was the right call or if it matters, but that was my logic.

desktop-exporter/app/routes/trace-view.tsx
CtrlSpice: I did it! 😁

jmorrell (approved):
DALL·E 2023-01-05 15 14 25 - A cute corgi saying _Nice Job!_

PR #51, opened 2023-01-03
fix the list/key debacle

CtrlSpice:
By wrapping all the duration labels except the last one in `ListItem`, then adding  spacer followed by the last duration label.
This way, there's no weird wrapping behaviour on resize.


jmorrell (approved):
DALL·E 2023-01-03 16 25 36 - A corgi with a rainbow unicorn horn, watercolor

PR #52, opened 2023-01-08
Handle orphan spans in WaterfallView

CtrlSpice:
Here, we write our own simplified implementation of [performant-array-to-tree](https://github.com/philipstanislaus/performant-array-to-tree) and modify it to handle incomplete traces where a span's parent is missing.

Note: I have to give further thought about how to visually represent the missing span (not clickable? empty detail view? different colour maybe?) so these things will be handled in another PR.

Also, I'm pretty sure I deleted all the batman jokes from the code before committing, but if not I'm sorry and let me know. 

missingParent

### DIFF
--- a/desktop-exporter/app/components/waterfall-view/waterfall-row.tsx
+++ b/desktop-exporter/app/components/waterfall-view/waterfall-row.tsx
@@ -25,7 +25,7 @@ export function WaterfallRow({ index, style, data }: WaterfallRowProps) {
     selectedSpanID,
     setSelectedSpanID,
   } = data;
-  let span = orderedSpans[index].span;
+  let { spanID, spanData } = orderedSpans[index];
   let spanDepth = orderedSpans[index].metadata.depth;
 
   // Set the background colour to make the list striped.
@@ -34,7 +34,7 @@ export function WaterfallRow({ index, style, data }: WaterfallRowProps) {
   let selectedColour = useColorModeValue("pink.50", "pink.900");
 
   //Set the style for the selected item
-  if (!!selectedSpanID && selectedSpanID === span.spanID) {
+  if (!!selectedSpanID && selectedSpanID === spanID) {
     backgroundColour = selectedColour;
   }
 
@@ -43,17 +43,22 @@ export function WaterfallRow({ index, style, data }: WaterfallRowProps) {
 
   // Add zero-width space after forward slashes, dashes, and dots
   // to indicate line breaking opportunity
-  let nameLabel = span.name
-    .replaceAll("/", "/\u200B")
-    .replaceAll("-", "-\u200B")
-    .replaceAll(".", ".\u200B");
+  let nameLabel = spanData
+    ? spanData.name
+        .replaceAll("/", "/\u200B")
+        .replaceAll("-", "-\u200B")
+        .replaceAll(".", ".\u200B")
+    : "missing span";
 
+  let resourceLabel = spanData
+    ? spanData.resource.attributes["service.name"]
+    : "";
   return (
     <Flex
       style={style}
       bgColor={backgroundColour}
       paddingLeft={`${paddingLeft}px`}
-      onClick={() => setSelectedSpanID(span.spanID)}
+      onClick={() => setSelectedSpanID(spanID)}
     >
       <Flex
         width={spanNameColumnWidth - paddingLeft}
@@ -72,7 +77,7 @@ export function WaterfallRow({ index, style, data }: WaterfallRowProps) {
         alignItems="center"
         paddingStart={3}
       >
-        <Text fontSize="sm">{span.resource.attributes["service.name"]}</Text>
+        <Text fontSize="sm">{}</Text>
       </Flex>
       <Spacer />
     </Flex>
--- a/desktop-exporter/app/components/waterfall-view/waterfall-row.tsx
+++ b/desktop-exporter/app/components/waterfall-view/waterfall-row.tsx
@@ -1,7 +1,8 @@
 import React from "react";
 import { Text, Flex, Spacer, useColorModeValue } from "@chakra-ui/react";
+import { WarningTwoIcon } from "@chakra-ui/icons";
 
-import { SpanWithUIData } from "../../types/metadata-types";
+import { SpanDataStatus, SpanWithUIData } from "../../types/metadata-types";
 
 type WaterfallRowData = {
   orderedSpans: SpanWithUIData[];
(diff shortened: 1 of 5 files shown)


desktop-exporter/app/types/metadata-types.ts
```diff
@@ -5,6 +5,7 @@ export type SpanMetaData = {
 };
 
 export type SpanWithMetadata = {
-  span: SpanData;
+  spanID: string;
+  spanData: SpanData | null;
   metadata: SpanMetaData;
 };
```
jmorrell, line 11: Another way to handle this might be a [discriminated union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) where you can have multiple "shapes" based on the value of a string literal type. In branches where you have already made sure `status === "present"` you should be able to access `.span` without having to guard against it not being there.

```typescript
type SpanWithMetaData = {
  status: "present";
  spanID: string;
  span: SpanData;
  metadata: SpanMetaData;
} | {
  status: "missing";
  spanID: string;
  metadata: SpanMetaData;
}

let a: SpanWithMetaData = {
  status: "missing",
  spanID: "1234",
  metadata: meta,
}
```

or you could put the `spanID` in `SpanMetaData` (maybe `SpanUIData`?)

```typescript
export type SpanUIData = {
  depth: number;
  spanID: string;
};

type SpanWithUIData = {
  status: "present";
  span: SpanData;
  metadata: SpanUIData;
} | {
  status: "missing";
  metadata: SpanUIData;
}
```

desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -0,0 +1,48 @@
...
+      if (!lookup[parentSpanID]) {
+        lookup[parentSpanID] = {
+          span: { spanID: parentSpanID, spanData: null },
+          children: [],
+        };
+        missingSpanIDs.add(parentSpanID);
+      }
+      lookup[parentSpanID].children.push(treeItem);
+    }
+  }
```
jmorrell, line 41: This took me longer to trace through than I would like to admit

jmorrell (approved):
DALL·E 2023-01-07 20 31 11 - a chibi corgi, 3d art

I have a suggestion for a different way to type this, but this would be fine if you wanted to ship it as-is. Great job! 💯

desktop-exporter/app/types/metadata-types.ts
```diff
@@ -1,11 +1,22 @@
 import { SpanData } from "./api-types";
 
-export type SpanMetaData = {
-  depth: number;
-};
+export enum SpanDataStatus {
+  missing,
+  present,
+}
```
jmorrell, line 6: I forgot that TypeScript added enums like this to the language. Using it like this can create issues while debugging since logging out the object will log `0` and `1` for `status` instead of something human-readable.

The solution to that problem is to add string values to your enum like this:

```typescript
export enum SpanDataStatus {
  missing = "missing",
  present = "present",
}
```

desktop-exporter/app/components/waterfall-view/waterfall-row.tsx
```diff
@@ -25,40 +25,47 @@ export function WaterfallRow({ index, style, data }: WaterfallRowProps) {
...
-  let nameLabel = spanData
-    ? spanData.name
-        .replaceAll("/", "/\u200B")
-        .replaceAll("-", "-\u200B")
-        .replaceAll(".", ".\u200B")
-    : "missing span";
+  let nameLabel;
+  let resourceLabel;
+
+  if (span.status) {
```
jmorrell, line 44: I was a little confused how this is working since `span.status` is not nullable, but I think this is currently relying on `SpanDataStatus.missing` being `0`. If we change it to string values (see other comment), this will break, so let's do an explicit check here:

```typescript
if (span.status === SpanDataStatus.present) {
```

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -20,62 +20,31 @@ export async function traceLoader({ params }: any) {
...
+    return orderedSpans[0].metadata.spanID;
+  });
 
   // if we get a new trace because the route changed, reset the selected span
   React.useEffect(() => {
-    setSelectedSpanID(orderedSpans[0].spanID);
+    setSelectedSpanID(
+      orderedSpans[0].status
+        ? orderedSpans[0].metadata.spanID
+        : orderedSpans[1].metadata.spanID,
```
jmorrell, line 46: Just to check, will we ever have a situation where both of these are missing? 🤔 I think no?

To be safer we might try something like:

```typescript
orderedSpans.find(span => span.status === SpanDataStatus.present).metadata.spanID
```

Except Typescript is going to complain that the `find` result could be `null` 🤔 

Maybe just leave it as-is

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -20,62 +20,31 @@ export async function traceLoader({ params }: any) {
...
-    orderedSpans.length ? orderedSpans[0].spanID : "",
-  );
+    return orderedSpans[0].metadata.spanID;
+  });
 
   // if we get a new trace because the route changed, reset the selected span
   React.useEffect(() => {
-    setSelectedSpanID(orderedSpans[0].spanID);
+    setSelectedSpanID(
+      orderedSpans[0].status
```
jmorrell, line 44:
```typescript
orderedSpans[0].status === SpanDataStatus.present
```

desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -1,45 +1,73 @@
...
 
     let treeItem = lookup[spanID];
 
+    // If the current span has no parentSpanID, add it to the rootItems
     if (!parentSpanID) {
       rootItems.push(treeItem);
     } else {
+      // If the current span's parentSpanID is not in the lookup structure yet, add it
+      // as a missing/incomplete span, to be updated in a subsequent loop if found
+      // (See comment on previous block of code - line 33ish)
```
jmorrell, line 54: Great job updating and documenting the changes to this algorithm 🙌

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -111,4 +80,59 @@ export default function TraceView() {
       </GridItem>
     </Grid>
   );
+}
+
+function orderSpans(spanTree: TreeItem[]): SpanWithUIData[] {
```
jmorrell, line 85: A comment here would be great. Something like:

```
// Do a depth-first traverse of the generated tree, re-flattening it out into an array
// ordering each set of children by start time and capturing information about the 
// depth of each span so we can render it correctly
//
// In the case that we are missing spans in the tree, orphaned subtrees will have a
// phantom parent span
```

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -111,4 +80,59 @@ export default function TraceView() {
...
+  let orderedSpans: SpanWithUIData[] = [];
+
+  for (let root of spanTree) {
+    let stack = [
+      {
+        treeItem: root,
+        depth: 0,
+      },
+    ];
```
jmorrell, line 94: We are sorting each set of children, but not the set of root nodes we are starting with. We might wish to sort `spanTree` before we enter this loop so that the root span is first and all missing spans come after.

desktop-exporter/app/routes/trace-view.tsx
```diff
@@ -111,4 +80,59 @@ export default function TraceView() {
...
+      } else {
+        orderedSpans.push({
+          status: SpanDataStatus.missing,
+          metadata: { depth: depth, spanID: treeItem.spanID },
+        });
+      }
+
+      treeItem.children
+        .sort((a, b) => {
+          if (a.status && b.status) {
```
jmorrell, line 118: if (a.status == SpanDataStatus.present && b.status == SpanDataStatus.present) {

jmorrell (approved):
Great job navigating some tricky typing!

DALL·E 2023-01-08 20 50 48 - cute kawaii sad orange cat holding a union sign, 3d render

desktop-exporter/app/routes/trace-view.tsx
CtrlSpice: The array-to-tree implementation takes care of that, as any spans left in the missing set are added to the rootSpans array only after all the spans have been processed, and the for loop exits. I added a comment both in `trace-view` and `array-to-tree`

jmorrell (approved):
DALL·E 2023-01-10 18 57 31

PR #53, opened 2023-01-11
Draw duration bars for each span

CtrlSpice:
They still need time indicators, labels, and event dots but look at them! Look. At. Them!

durationBars
durationBarsDay

### DIFF
--- a/desktop-exporter/app/components/waterfall-view/duration-bar.tsx
+++ b/desktop-exporter/app/components/waterfall-view/duration-bar.tsx
@@ -1,0 +1,45 @@
+import React from "react";
+import { Box, Flex, useColorModeValue } from "@chakra-ui/react";
+
+import { NsFromString, TraceTimeAttributes } from "../../utils/duration";
+import { EventData } from "../../types/api-types";
+
+type DurationBarProps = {
+  events: EventData[];
+  traceTimeAttributes: TraceTimeAttributes;
+  spanStartTimestamp: string;
+  spanEndTimestamp: string;
+};
+
+export function DurationBar(props: DurationBarProps) {
+  let durationBarColour = useColorModeValue("cyan.800", "cyan.700");
+
+  let { traceStartTimeNS, traceDurationNS } = props.traceTimeAttributes;
+  let spanStartTimeNs = NsFromString(props.spanStartTimestamp);
+  let spanEndTimeNs = NsFromString(props.spanEndTimestamp);
+
+  let offsetStart = Math.floor(
+    ((spanStartTimeNs - traceStartTimeNS) / traceDurationNS) * 100,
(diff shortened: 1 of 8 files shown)


desktop-exporter/app/utils/duration.ts
```diff
@@ -1,28 +1,55 @@
...
+    milliseconds * 1e6 + Timestamp.fromString(timestampString).getNano();
+  return nanoseconds;
 }
 
-export function getSpanDurationString(
-  startTimestamp: string,
-  endTimestamp: string,
-) {
-  let durationNs = getSpanDuration(startTimestamp, endTimestamp);
+export function DurationNs(startTimestamp: string, endTimestamp: string) {
```
jmorrell, line 45: Capitalizing these function names is a bit confusing because then the look like React components or a Class when you are using them. Could we change them back?

desktop-exporter/app/components/waterfall-view/duration-bar.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import { Box, Flex, useColorModeValue } from "@chakra-ui/react";
+
+import { NsFromString, TraceTimeAttributes } from "../../utils/duration";
+import { EventData } from "../../types/api-types";
+
+type DurationBarProps = {
+  events: EventData[];
```
jmorrell, line 8: You probably want to pass the whole `spanData` down instead of just the events since you will need it if you ever want to color the bar based on span values.

desktop-exporter/app/utils/duration.ts
```diff
@@ -1,28 +1,55 @@
...
-
-    return endTimeNs - startTimeNs;
-  } catch (error) {
-    return null;
+export type TraceTimeAttributes = {
+  traceStartTimeNS: number;
+  traceDurationNS: number;
+};
+
+export function getTraceTimeAttributes(spans: SpanData[]): TraceTimeAttributes {
```
jmorrell, line 9: This name is a bit odd. `calculateTraceDuration` maybe?

jmorrell (approved):
THIS LOOKS SO GOOD!
DALL·E 2023-01-10 19 16 30 - a corgi painting a waterfall, digital art

desktop-exporter/app/utils/duration.ts
CtrlSpice: Yes! I went ahead and overthinked this, hard. Going back to getThingXYZed.

desktop-exporter/app/utils/duration.ts
CtrlSpice: I went with calculateTraceTiming because I need the start and the duration.

PR #55, opened 2023-01-13
Draw event circles on the duration bar -draft

CtrlSpice:
I need a bit of design help for this one, especially when the duration labels and the event indicators overlap and become illegible. And just about their look in general because I don't like them but don't know how to fix them.

eventIndicatorsDraft

### DIFF
--- a/desktop-exporter/app/components/waterfall-view/duration-bar.tsx
+++ b/desktop-exporter/app/components/waterfall-view/duration-bar.tsx
@@ -15,7 +15,49 @@ import {
   getNsFromString,
   TraceTiming,
 } from "../../utils/duration";
-import { SpanData } from "../../types/api-types";
+import { EventData, SpanData } from "../../types/api-types";
+
+type EventDotsListProps = {
+  events: EventData[];
+  spanStartTimeNs: number;
+  spanEndTimeNs: number;
+};
+
+function EventDotsList(props: EventDotsListProps) {
+  let { events, spanStartTimeNs, spanEndTimeNs } = props;
+
+  let eventDotsList = events.map((eventData) => {
+    let eventName = eventData.name;
+    let eventTimeNs = getNsFromString(eventData.timestamp);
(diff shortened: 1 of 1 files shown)


jmorrell:
I think this looks okay? 🤔  Maybe we could move the labels a little farther in since the events are more likely at the very beginning of the span?

CtrlSpice:
Added some tooltips!

desktop-exporter/app/components/waterfall-view/duration-bar.tsx
```diff
@@ -51,12 +59,41 @@ export function DurationBar(props: DurationBarProps) {
...
+            borderColor="cyan.800"
+            position="absolute"
+            left={`${eventOffsetPercent}%`}
+            transformOrigin="center"
+            transform="translate(-50%)"
+          />
+        </Tooltip>
+      </li>
+    );
+  });
```
jmorrell, line 91: Consider breaking this out as an `EventCircleList` component

desktop-exporter/app/components/waterfall-view/duration-bar.tsx
```diff
@@ -51,12 +59,41 @@ export function DurationBar(props: DurationBarProps) {
...
+  let eventCircles = props.spanData.events.map((eventData, index) => {
+    let eventName = eventData.name;
+    let eventTimeNs = getNsFromString(eventData.timestamp);
+    let spanDurationNS = spanEndTimeNs - spanStartTimeNs;
+    let eventOffsetPercent = Math.floor(
+      ((eventTimeNs - spanStartTimeNs) / spanDurationNS) * 100,
+    );
+
+    return (
+      <li key={index}>
```
jmorrell, line 72: `key={index}` is a bit of an anti-pattern since a new list with different values would have the same keys. Can you pull out an id here? or `${eventName}-${eventTimeNs}`?

jmorrell (approved):
DALL·E 2023-01-14 18 32 13 - a corgi floating among mathematical circles in a dreamy pattern

PR #56, opened 2023-01-16
Update TraceSummary on the back-end in preparation for sidebar rework

CtrlSpice:
TraceSummary is now structured as follows:

- HasRootSpan (bool)
- RootServiceName (string)
- RootName (string)
- RootStartTime (time.Time)
- RootEndTime  (time.Time)
- SpanCount  (uint32)
- TraceID  (string)

I covered the possibility of a partial trace with no root span, and will reflect this on the front-end.

Tests and types were updated accordingly.

I opted for a start and end time for the root span rather than durationMS because we already have a good setup on the front-end to calculate duration and output it in the most human-readable time unit depending on its value, so it doesn't make sense to duplicate the functionality and limit it to milliseconds.

### DIFF
--- a/desktop-exporter/trace_data.go
+++ b/desktop-exporter/trace_data.go
@@ -1,40 +1,38 @@
 package desktopexporter
 
-import (
-	"time"
-)
+import "time"
 
-func (trace *TraceData) GetTraceSummary() (TraceSummary, error) {
-	duration, err := trace.getTraceDuration()
-	if err != nil {
-		return TraceSummary{}, err
+func (trace *TraceData) GetTraceSummary() (TraceSummary) {
+	rootSpan, err := trace.getRootSpan()
+	
+	if err == ErrMissingRootSpan {
+		return TraceSummary{
+			HasRootSpan:     false,
+			RootServiceName: "",
+			RootName:        "",
+			RootStartTime:   time.Time{},
+			RootEndTime:	 time.Time{},
+			SpanCount:       uint32(len(trace.Spans)),
+			TraceID:         trace.TraceID,
+		}
 	}
 
 	return TraceSummary{
-		TraceID:    trace.TraceID,
-		SpanCount:  uint32(len(trace.Spans)),
-		DurationMS: duration.Milliseconds(),
-	}, nil
-
+		HasRootSpan:     true,
+		RootServiceName: rootSpan.Attributes["service.name"].(string),
+		RootName:        rootSpan.Name,
+		RootStartTime:   rootSpan.StartTime,
+		RootEndTime:	 rootSpan.EndTime,
+		SpanCount:       uint32(len(trace.Spans)),
+		TraceID:         trace.TraceID,
+	}
 }
 
-func (trace *TraceData) getTraceDuration() (time.Duration, error) {
-	if len(trace.Spans) < 1 {
-		return 0, ErrEmptySpansSlice
-	}
-
-	// Determine the total duration of the trace
-	traceStartTime := trace.Spans[0].StartTime
-	traceEndTime := trace.Spans[0].EndTime
-	for i := 1; i < len(trace.Spans); i++ {
-
-		if trace.Spans[i].StartTime.Before(traceStartTime) {
-			traceStartTime = trace.Spans[i].StartTime
-		}
-
-		if trace.Spans[i].EndTime.After(traceEndTime) {
-			traceEndTime = trace.Spans[i].EndTime
+func (trace *TraceData) getRootSpan()(SpanData, error){
+	for i := 0; i < len(trace.Spans); i++ {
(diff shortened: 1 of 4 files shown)


CtrlSpice:
Fixed my logic errors, my naming errors, AND my test errors. 
Also, set the maximum number of traces held in memory to 10K for now. We can make it configurable in the future, but I needed a sane number to hardcode for the time being.
Now ready for review.

desktop-exporter/trace_data.go
```diff
@@ -4,37 +4,37 @@ import (
...
-		DurationMS: duration.Milliseconds(),
-	}, nil
-
-}
-
-func (trace *TraceData) getTraceDuration() (time.Duration, error) {
-	if len(trace.Spans) < 1 {
-		return 0, ErrEmptySpansSlice
+		HasRootSpan:     true,
+		RootServiceName: rootSpan.Resource.Attributes["service.name"].(string),
```
jmorrell, line 24: What happens when this isn't present? What happens if this is somehow set to a bool or float? 🤔

jmorrell: Consider factoring this out into a method on `SpanData` https://github.com/CtrlSpice/otel-desktop-viewer/blob/main/desktop-exporter/types.go#L39

which will make easy to test in isolation and simplifies this line to:

```go
RootServiceName: rootSpan.ServiceName()
```

desktop-exporter/trace_data_test.go
```diff
@@ -18,22 +18,56 @@ func TestGetTraceSummary(t *testing.T) {
...
+	assert.Equal(t, time.Date(2020, 10, 21, 7, 10, 2, 300, time.UTC), summary.RootEndTime)
+	assert.Equal(t, uint32(spansPerTrace), summary.SpanCount)
+	assert.Equal(t, "1", summary.TraceID)
+}
+
+func TestGetTraceSummaryMissingRootSpan(t *testing.T) {
+	maxQueueLength := 1
+	spansPerTrace := 3
+
+	traces := testdata.GenerateOTLPPayload(1, 1, maxQueueLength*spansPerTrace)
```
jmorrell, line 52: I didn't realize `GenerateOTLPPayload` doesn't create any root spans. It would be good to add that as a comment

```suggestion
	// All spans from testdata.GenerateOTLPPayload have the same parent span ID
	traces := testdata.GenerateOTLPPayload(1, 1, maxQueueLength*spansPerTrace)
```

desktop-exporter/trace_data.go
```diff
@@ -4,37 +4,37 @@ import (
 	"time"
 )
 
-func (trace *TraceData) GetTraceSummary() (TraceSummary, error) {
-	duration, err := trace.getTraceDuration()
-	if err != nil {
-		return TraceSummary{}, err
+func (trace *TraceData) GetTraceSummary() (TraceSummary) {
```
jmorrell, line 7: The refactor here is 💯

jmorrell (approved):
I wasn't sure which one I liked better, so here's a 2-fer

DALL·E 2023-01-16 14 15 02 - A corgi eating an enormous pile of tacos, digital art
DALL·E 2023-01-16 14 14 58 - A corgi eating an enormous pile of tacos, digital art

desktop-exporter/types.go
```diff
@@ -8,15 +8,22 @@ import (
...
-	DurationMS int64  `json:"durationMS"`
+	HasRootSpan 	bool	`json:"hasRootSpan"`
+
+	RootServiceName string		`json:"rootServiceName"`
+	RootName 		string		`json:"rootName"`
+	RootStartTime 	time.Time	`json:"rootStartTime"`
+	RootEndTime   	time.Time	`json:"rootEndTime"`
+	
+	SpanCount  		uint32	`json:"spanCount"`
+	TraceID    		string	`json:"traceID"`
```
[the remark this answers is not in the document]
CtrlSpice: Yes, my ide stopped formatting go on save for some reason.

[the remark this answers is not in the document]
CtrlSpice: gofmt -w desktop-exporter/types.go appears to have fixed it.

desktop-exporter/trace_data.go
CtrlSpice: Hey, so it is my understanding that if it is somehow set to a float or a bool, or isn't present, there is an issue in the SDK because it's supposed to fall back to `unknown_service :` + `process.executable.name` or failing that, just `unknown_service`. I've factored it out though, and set it to warn about missing or invalid service names.

Here's the thing I read: https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service
I decided not to fall back on `unknown_service` in the app, since that could mask a problem for someone if they are using my thing to check if they've instrumented their code correctly? Does that sound right? I just output an empty string...

What do you think?

PR #62, opened 2023-01-22
Fix span status and kind post-update

CtrlSpice:
Span status and span kind are just passed as strings now, so we no longer need the enum mapping on the front-end 🥳


jmorrell (approved):
The best diff is a red diff 🔴 
otel-desktop-viewer 2023-01-21 21-08-39

DALL·E 2023-01-21 21 10 18 - corgi covered in red paint, digital art

PR #63, opened 2023-01-22
Remove hex conversions in favour of .String() for Trace/Span IDs

CtrlSpice:
So empty `parentSpanID`s no longer show up as a string of zeros and make it so that none of the traces appear to have a root span.


jmorrell (approved):
DALL·E 2023-01-21 21 46 44 - a corgi performing an exorcism, digital art

PR #64, opened 2023-01-22
Add go:embed and serve static files that way

CtrlSpice:
Also remove unused .css file from static, and fix build step in make file.


jmorrell (approved):
DALL·E 2023-01-22 15 49 10 - a corgi with raised fur because of static electricity

PR #65, opened 2023-01-23
change the config file to a string-o-yaml

CtrlSpice:
In order to make this as easy to install as possible.


jmorrell (approved):
Dalle is confused as to what a yaml is
DALL·E 2023-01-22 16 15 14 - yaml, digital art

PR #66, opened 2023-01-25
Fix css scrolling bug

CtrlSpice:
Who's got two thumbs and deleted the main.css file in the static dir instead of the app? Me. It was me. 🤦‍♀️ 

Anyway, updated main.css IN THE CORRECT directory this time which fixed the scrolling issue, and updated the title while I was at it.

### DIFF
--- a/desktop-exporter/static/index.html
+++ b/desktop-exporter/static/index.html
@@ -2,11 +2,11 @@
 <html>
 <head>
   <meta charset=utf-8>
-  <title>Title Here</title>
+  <title>Otel Desktop Viewer</title>
   <link rel="stylesheet" href="/main.css"/>
 </head>
 <body>
-  <div id="root"></div>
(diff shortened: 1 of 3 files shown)


desktop-exporter/static/index.html
```diff
@@ -2,11 +2,11 @@
 <html>
 <head>
   <meta charset=utf-8>
-  <title>Title Here</title>
+  <title>Otel Desktop Viewer</title>
```
jmorrell, line 5: Should we use "OpenTelemetry" here instead of "Otel"?

jmorrell (approved):
DALL·E 2023-01-25 11 46 39 - scrolling bug, digital art

desktop-exporter/static/index.html
CtrlSpice: Yes.

PR #67, opened 2023-01-25
display links panel in detail view

CtrlSpice:
With minimal functionality (read-only) and an under construction banner.

Please note: I temporarily commented out the go:embed functionality to make the rest of my front-end tasks easier. If there's a better way to go about this please let me know.

Also, there's screenshots:
 
links
link-day

### DIFF
--- a/desktop-exporter/app/components/detail-view/detail-view.tsx
+++ b/desktop-exporter/app/components/detail-view/detail-view.tsx
@@ -1,9 +1,11 @@
 import React from "react";
 import { Flex, Tab, TabList, TabPanels, Tabs } from "@chakra-ui/react";
 
+
 import { SpanData } from "../../types/api-types";
 import { FieldsPanel } from "./fields-panel";
 import { EventsPanel } from "./events-panel";
+import { LinksPanel } from "./links-panel";
 
 type DetailViewProps = {
   span: SpanData | undefined;
@@ -16,7 +18,6 @@ export function DetailView(props: DetailViewProps) {
   }
   let numEvents = span.events.length;
   let numLinks = span.links.length;
-
   return (
     <Flex
       grow="0"
@@ -43,6 +44,7 @@ export function DetailView(props: DetailViewProps) {
             events={span.events}
             spanStartTime={span.startTime}
           />
+          <LinksPanel links={span.links} />
         </TabPanels>
       </Tabs>
     </Flex>
--- a/desktop-exporter/app/components/detail-view/detail-view.tsx
+++ b/desktop-exporter/app/components/detail-view/detail-view.tsx
@@ -1,7 +1,6 @@
 import React from "react";
 import { Flex, Tab, TabList, TabPanels, Tabs } from "@chakra-ui/react";
 
-
 import { SpanData } from "../../types/api-types";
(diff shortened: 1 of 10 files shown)


desktop-exporter/app/components/detail-view/links-panel.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <SpanField
+              fieldName="dropped attributes count"
+              fieldValue={link.droppedAttributesCount}
+              hidden={!link.droppedAttributesCount}
+            />
+          </AccordionPanel>
+        </AccordionItem>
+      </li>
+    );
+  });
```
jmorrell, line 72: Once a `map` gets this large, I recommend breaking it out as a separate component

```jsx
linkListItems = links.map((link) => <LinkItem link={link}/>)
```

desktop-exporter/server.go
```diff
@@ -79,25 +77,26 @@ func traceIDHandler(store *TraceStore) func(http.ResponseWriter, *http.Request)
...
-            return
-	}
-	writer.Write(indexBytes)
+	http.ServeFile(writer, request, "./desktop-exporter/static/index.html")
+	// indexBytes, err := assets.ReadFile("static/index.html")
+	// if err != nil {
+	// 	writer.WriteHeader(http.StatusInternalServerError)
+    //         return
+	// }
+	// writer.Write(indexBytes)
```
jmorrell, line 86: You might do something like:

```go
if os.Getenv("SERVE_FROM_FS") == "true" {
    http.ServeFile(writer, request, "./desktop-exporter/static/index.html")
} else {
    indexBytes, err := assets.ReadFile("static/index.html")
    if err != nil {
        writer.WriteHeader(http.StatusInternalServerError)
        return
    }
    writer.Write(indexBytes)
}
```

so you can run during development like:

```
$ SERVE_FROM_FS=true go run ./...
```

jmorrell (approved):
DALL·E 2023-01-26 08 16 06 - a corgi with a warning light on its head, digital art

desktop-exporter/app/components/detail-view/links-panel.tsx
CtrlSpice: Done here and for the same issue in the events tab

desktop-exporter/server.go
CtrlSpice: Thank you!

desktop-exporter/app/components/detail-view/detail-view.tsx
```diff
@@ -1,9 +1,11 @@
 import React from "react";
 import { Flex, Tab, TabList, TabPanels, Tabs } from "@chakra-ui/react";
 
+
```
jmorrell, line 4: nit, extra line. Prettier should consolidate these I thought?

jmorrell (approved):
DALL·E 2023-01-16 14 15 02 - A corgi eating an enormous pile of tacos, digital art

desktop-exporter/app/components/detail-view/detail-view.tsx
CtrlSpice: It treats lines between imports as intentional for some reason. Not sure why. I'll be more careful.

PR #68, opened 2023-01-26
Launch browser window

CtrlSpice:
When we start the server, we open a browser window to http://localhost:8000/


jmorrell (approved):
There's some unintended cruft in this I think. Let's fix that up before merge
DALL·E 2023-01-26 08 20 19 - a corgi browsing the web on a laptop, digital art

jmorrell (approved):
Dalle won't let me log in so take this puppy instead: 🐶

PR #73, opened 2023-02-02
Provide keys for some list children

CtrlSpice:
Fixed a rendering bug in events- and links- views that I introduced earlier and failed to spot.

### DIFF
--- a/desktop-exporter/app/components/detail-view/events-panel.tsx
+++ b/desktop-exporter/app/components/detail-view/events-panel.tsx
@@ -26,7 +26,7 @@ function EventItem(props: EventItemProps) {
   let timeSinceSpanStart = getDurationNs(spanStartTime, event.timestamp);
   let durationString = getDurationString(timeSinceSpanStart);
   let eventAttributes = Object.entries(event.attributes).map(([key, value]) => (
-    <li key={key + value?.toString}>
+    <li key={key + value?.toString()}>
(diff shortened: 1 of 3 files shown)


desktop-exporter/app/components/detail-view/events-panel.tsx
```diff
@@ -26,7 +26,7 @@ function EventItem(props: EventItemProps) {
   let timeSinceSpanStart = getDurationNs(spanStartTime, event.timestamp);
   let durationString = getDurationString(timeSinceSpanStart);
   let eventAttributes = Object.entries(event.attributes).map(([key, value]) => (
-    <li key={key + value}>
+    <li key={key + value?.toString}>
```
jmorrell, line 29: I think `value?.toString` will either be null or a reference to a function?

```suggestion
    <li key={key + value?.toString()}>
```

jmorrell (changes requested):

desktop-exporter/app/components/detail-view/events-panel.tsx
jmorrell: Since `key` is always a string, I think value will be coerced to a string, though that may not play well with types

jmorrell (approved):
DALL·E 2023-02-03 16 52 39 - a corgi ironman, digital art

PR #74, opened 2023-02-04
Replace repeated refresh with polling in empty state

CtrlSpice:
App now polls trace store for number of traces every half-second and reloads when there's something there.

### DIFF
--- a/desktop-exporter/server.go
+++ b/desktop-exporter/server.go
@@ -25,6 +25,23 @@ type Server struct {
 	traceStore *TraceStore
 }
 
+func traceCountHandler(store *TraceStore) func(http.ResponseWriter, *http.Request){
+	return func(writer http.ResponseWriter, request *http.Request) {
+		numTraces := len(store.traceMap)
+		if numTraces > maxNumTraces {
+			numTraces = maxNumTraces
+		}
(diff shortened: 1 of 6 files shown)


jmorrell:
This is a much better experience! Do you think we could auto-show the most recent trace? or is that a further change?

https://user-images.githubusercontent.com/175496/216736399-d1a7237e-0390-4811-8cb8-da95ae1906ef.mp4

desktop-exporter/server.go
```diff
@@ -25,6 +25,23 @@ type Server struct {
 	traceStore *TraceStore
 }
 
+func traceCountHandler(store *TraceStore) func(http.ResponseWriter, *http.Request){
```
jmorrell, line 28: I'm not sure this needs a new handler. Could we re-use the trace summaries? Then we could redirect to the latest trace directly.

CtrlSpice: Yes! Good call!

CtrlSpice:
Auto-showing the most recent trace is the thing I'm doing in my next PR 😄

jmorrell (approved):
DALL·E 2023-02-04 17 23 00 - a corgi floating in vast emptiness, digital art

Note that this needs to be rebased before it is merged.

PR #79, opened 2023-02-09
Fix boolean attributes display bug

CtrlSpice:
- Fixed the bug where boolean attributes were displayed as an empty string
- Also fixed the bug where the type of the attribute was checked after conversion and was always `string`

boolean-attributes

### DIFF
--- a/desktop-exporter/app/components/detail-view/span-field.tsx
+++ b/desktop-exporter/app/components/detail-view/span-field.tsx
@@ -21,15 +21,24 @@ export function SpanField(props: SpanFieldProps) {
   if (hidden) {
     return null;
   }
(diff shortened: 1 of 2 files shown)


desktop-exporter/app/components/detail-view/span-field.tsx
```diff
@@ -21,15 +21,24 @@ export function SpanField(props: SpanFieldProps) {
   if (hidden) {
     return null;
   }
+  let typeOfFieldValue = typeof fieldValue;
```
jmorrell, line 24: 💯

jmorrell (approved):
DALL·E 2023-02-09 13 42 59 - a corgi dressed as a ghost saying boolean, illustration

PR #80, opened 2023-02-12
Add small delays

CtrlSpice:
- before launching the browser to let the server start and avoid a 404 as the first user experience (250ms)
- before reloading the window when we receive traces in order to let more than one payload come through and to have more complete traces to look at (500ms)


desktop-exporter/app/components/empty-state-view/empty-state-view.tsx
```diff
@@ -65,7 +65,9 @@ async function pollTraceCount() {
   } else {
     let { traceSummaries } = (await response.json()) as TraceSummaries;
     if (traceSummaries.length > 0) {
-      window.location.reload();
+      setTimeout(() => {
+        window.location.reload();
+      }, 500);
```
jmorrell, line 70: You can write this a little simpler as

```
setTimeout(window.location.reload, 500);
```

jmorrell (approved):

PR #81, opened 2023-02-12
Default to dark mode

CtrlSpice:
when the application is first launched because dark mode is just prettier


jmorrell (approved):
DALL·E 2023-02-12 15 16 43 - a corgi reading ghost stories under a blanket in the dark, kids cartoon

PR #82, opened 2023-02-12
fix delay bug

CtrlSpice:
fix bug introduced by passing window.location.reload function directly to setTimeout


jmorrell (approved):
My bad 😅 

DALL·E 2023-02-12 15 48 43 - a corgi and bug are best friends, floating heart, childrens cartoon

Issue #92, opened 2023-02-19
Duration label text wrapping


CtrlSpice:
![wrap](https://user-images.githubusercontent.com/56372758/219909471-b63fba12-a30b-4e8d-b277-b2cde8536b76.png)
The duration and unit of measurement should be inline.

PR #93, opened 2023-02-19
Fix wrapping on whitespace in duration label

CtrlSpice:
Fixes issue #92 
-Added missing `whiteSpace="nowrap` to the duration label. It is no longer misbehaving.
inline-duration-label


jmorrell (approved):
DALL·E 2023-02-19 13 20 58 - a corgi tangled up in a sheet, digital art

PR #95, opened 2023-02-20
Sort missing spans at root level

### LINKED ISSUE #94: Sort missing spans
Right now, when we have an incomplete trace, missing spans at root level are added after the root span in no particular order.
To keep the easy visual parsing of the waterfall view, they should be sorted by the start time of their earliest child span.

Here's a view of the current situation:
unsorted-missing-spans

CtrlSpice:
Closes #94 
- Missing spans are now ordered at root level by the starTime of their children, from earliest to last, preserving the visual integrity of the waterfall view and making the data easier to parse at a glance. 
sorted-missing-spans
- Could use refactoring advice re: `array-to-tree` as it is getting quite verbose 🙏

### DIFF
--- a/desktop-exporter/app/utils/array-to-tree.ts
+++ b/desktop-exporter/app/utils/array-to-tree.ts
@@ -1,5 +1,6 @@
 import { SpanData } from "../types/api-types";
 import { SpanDataStatus } from "../types/ui-types";
+import { getNsFromString } from "./duration";
 
 export type TreeItem =
   | {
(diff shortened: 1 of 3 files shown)


desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -67,9 +68,74 @@ export function arrayToTree(spans: SpanData[]): TreeItem[] {
...
 
-  // In order to handle incomplete traces, the missing spans get appended to the end of the rootItems array
-  // This way we make sure that the root span is added first (if present), followed by any missing spans
-  for (let spanID of missingSpanIDs) {
+  // To handle incomplete traces:
+  // 1. Sort the missing spans by the earliest start time of their children
+  // 2. Appended sorted spand to the end of the rootItems array
+  // This way we make sure that the root span is added first (if present),
+  // and preserve the visual clarity of the waterfall view
+  let missingSpans = Array.from(missingSpanIDs).sort((a: string, b: string) => {
```
jmorrell, line 76: One option to simplify this logic is more precise types. Something like:

```typescript
export type RootTreeItem =
  | {
      status: SpanDataStatus.present;
      spanData: SpanData;
    children: TreeItem[];
    }
  | {
      status: SpanDataStatus.missing;
      spanID: string;
    children: TreeItem[];
    };

export type TreeItem = {
  spanData: SpanData;
  children: TreeItem[];
}
```

So that missing spans can only happen in the root of an `RootTreeItem[]`

Baring that, a helper function could help here:

```typescript
function getEarliestStartTime(children: TreeItem[]): number {
  if (children.length == 0) {
    throw new Error(
      "Unexpected type: A 'missing' parent span appears to have no children.",
    );
  }

  let startTimes = children.map((treeItem) => {
    if (treeItem.status === SpanDataStatus.missing) {
      throw new Error(
        "Unexpected type: A child of a 'missing' parent span appears to have no SpanData.",
      );
    }
    // getNsFromString returns `any` instead of `number`
    // this could also be fixed
    return getNsFromString(treeItem.spanData.startTime) as number;
  });

  return Math.min(...startTimes)
}
```

CtrlSpice: TFW you forget Math.min exists. It's much nicer now!

desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -67,11 +68,47 @@ export function arrayToTree(spans: SpanData[]): TreeItem[] {
...
+  // This way we make sure that the root span is added first (if present),
+  // and preserve the visual clarity of the waterfall view
+  let missingIDsArray = Array.from(missingSpanIDs).sort(
+    (a: string, b: string) => {
+      let earliestStartTimeA = getEarliestStartTime(lookup[a].children);
+      let earliestStartTimeB = getEarliestStartTime(lookup[b].children);
+
+      return earliestStartTimeA - earliestStartTimeB;
+    },
+  );
```
jmorrell, line 83: Much improved 💯

jmorrell (approved):
DALL·E 2023-02-20 14 43 27 - A line or corgis sorted smallest to largest, digital art

PR #96, opened 2023-02-20
Refactor tree node types

jmorrell:
I nerd-sniped myself with [this comment](https://github.com/CtrlSpice/otel-desktop-viewer/pull/95#discussion_r1112302232) and wanted to see if I could refactor the types this way.

This splits the type `TreeItem` into 2 different named types: `TreeItem` and `MissingTreeItem` which get unified as `RootTreeItem`. 

`RootTreeItem` can be either a `TreeItem` (present) or a `MissingTreeItem` (missing), but in either case it has to have `TreeItem` as children.

This enforces in the type system that missing spans can *only* appear at the root of the tree

https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions

### DIFF
--- a/desktop-exporter/app/utils/array-to-tree.ts
+++ b/desktop-exporter/app/utils/array-to-tree.ts
@@ -2,21 +2,25 @@ import { SpanData } from "../types/api-types";
 import { SpanDataStatus } from "../types/ui-types";
 import { getNsFromString } from "./duration";
 
-export type TreeItem =
-  | {
-      status: SpanDataStatus.present;
-      spanData: SpanData;
-      children: TreeItem[];
-    }
-  | {
-      status: SpanDataStatus.missing;
-      spanID: string;
-      children: TreeItem[];
-    };
+export type TreeItem = {
+  status: SpanDataStatus.present;
+  spanData: SpanData;
+  children: TreeItem[];
+};
 
-export function arrayToTree(spans: SpanData[]): TreeItem[] {
-  let rootItems: TreeItem[] = [];
-  let lookup: { [spanID: string]: TreeItem } = {};
+export type MissingTreeItem = {
+  status: SpanDataStatus.missing;
+  spanID: string;
+  children: TreeItem[];
+};
+
+export type RootTreeItem =
+  | TreeItem
+  | MissingTreeItem;
+
+export function arrayToTree(spans: SpanData[]): RootTreeItem[] {
+  let rootItems: RootTreeItem[] = [];
+  let lookup: { [spanID: string]: RootTreeItem } = {};
   let missingSpanIDs: Set<string> = new Set();
 
   for (let spanData of spans) {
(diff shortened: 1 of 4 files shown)


desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -99,14 +106,6 @@ function getEarliestStartTime(children: TreeItem[]): number {
...
 
   let startTimes = children.map((treeItem) => {
-    if (treeItem.status === SpanDataStatus.missing) {
-      throw new Error(
-        // This should also happen in this implementation, since that the child span
-        // must have SpanData (minimally a parentSpanID) in order for the parent span to be created.
-        "Unexpected type: A child of a 'missing' parent span appears to have no SpanData.",
-      );
-    }
-
```
jmorrell, line 109: Did I make the types more complicated just to get rid of this assertion? Yes... yes I did

desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -31,21 +35,24 @@ export function arrayToTree(spans: SpanData[]): TreeItem[] {
...
 
+    let treeItem = lookup[spanID];
+
     // Note A:
     // If the span has been added to the lookup structure as a missing/incomplete parent
     // on a previous pass (see Note B), update it and mark it present, and remove it from the missing set.
-    if (lookup[spanID].status === SpanDataStatus.missing) {
-      let children = lookup[spanID].children;
-      lookup[spanID] = {
+    if (treeItem.status === SpanDataStatus.missing) {
```
jmorrell, line 43: Instead of always referring to `lookup[spanID]`, if we reference `treeItem` here then the type system can follow that we are making assertions against it and track the flow

desktop-exporter/app/utils/array-to-tree.ts
```diff
@@ -31,21 +35,24 @@ export function arrayToTree(spans: SpanData[]): TreeItem[] {
...
+      let children = treeItem.children;
+      // Re-assign treeItem as a TreeItem type so that after this if statement,
+      // the type system knows that treeItem can only be a TreeItem
+      treeItem = {
         status: SpanDataStatus.present,
         spanData: spanData,
         children: children,
       };
+      lookup[spanID] = treeItem
```
CtrlSpice, line 52: Would it be possible to please add a comment here, saying that you are overwriting the tree item after you set it? There's just a lot going on in this function, in a weird order, and I will 100% forget about it the next time I look at it.

CtrlSpice (approved):
Awesome! Thank you! ❤️ 
branch

desktop-exporter/app/utils/array-to-tree.ts
jmorrell: Done!

PR #98, opened 2023-02-22
Replace default collector command line args with custom ones  

### LINKED ISSUE #85: Allow port configuration via command line arguments
Currently we have hard-coded all 3 ports that we use: otlp-grpc, otlp-http, and the http port used by the exporter. These should all be configurable via the command line.

This will entail needing to pass configuration down to the exporter, which might need to be tackled in a separate PR.

### LINKED ISSUE #87: Add CLI output for —help
CLIs should be helpful and explain how they are used. Might want to bring in a CLI argument parsing library for this.

CtrlSpice:
Closes #85, closes #87

- Allows us to easily set http, grpc, and browser endpoints via the command line
- Replaces the default collector commands, since we are hardcoding our `ConfigProvider` settings anyway
- Works on Windows!


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
jmorrell, line 1: you are `.gitignore`-ing this, but even so, you still need to remove it from your git history.

```shell
git rm .vscode/settings.json
```

main.go
```diff
@@ -2,30 +2,59 @@ package main
...
+		Version:     "0.1.1",
+	}
+
+	if err := run(info); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func runInteractive(buildInfo component.BuildInfo) error {
+	// cmd := otelcol.NewCommand(params)
```
jmorrell, line 27: nit: delete unused code

main_windows.go
```diff
@@ -40,9 +43,57 @@ func checkUseInteractiveMode() (bool, error) {
...
-func runService(params otelcol.CollectorSettings) error {
+func runService(buildInfo component.BuildInfo) error {
+	configContents := `yaml:
+receivers:
+  otlp:
+    protocols:
+      http:
+        endpoint: localhost:4318
+      grpc:
+        endpoint: localhost:4317
```
jmorrell, line 54: I'm unfamiliar with running things as a service on windows, but do we want to be able to customize ports there too? or is hard-coding them in this code-path an intentional choice?

CtrlSpice: It's intentional to the best of my understanding, and I am sure someone will correct me if I got it wrong, but command is not run on this path on the autogenerated code, so I went based on that,

PR #99, opened 2023-02-22
Allow configuration of the port on which we expose our frontend

CtrlSpice:
- Allow configuration of the port on which we expose our data via the yaml string defined in `main.go`
- This is done in preparation for allowing this port to be configured via command line

Look at Lulu, happily chilling on port 8880:
ConfigureFrontendPort

### DIFF
--- a/desktop-exporter/config.go
+++ b/desktop-exporter/config.go
@@ -8,7 +8,7 @@ import (
 
 // Config defines configuration for logging exporter.
 type Config struct {
-	// Endpoint defines where we expose our frontend.
+	// Endpoint defines where we serve our frontend app
 	Endpoint string `mapstructure:"endpoint"`
 }
 
--- a/desktop-exporter/config.go
+++ b/desktop-exporter/config.go
@@ -1,16 +1,24 @@
 package desktopexporter
 
 import (
+	"fmt"
+
 	"go.opentelemetry.io/collector/component"
 )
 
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -25,6 +25,7 @@ processors:
 
 exporters:
   desktop:
+    endpoint: localhost:8880
```
jmorrell, line 28: We probably don't want to change away from `localhost:8000` as the default

desktop-exporter/config.go
```diff
@@ -1,16 +1,24 @@
...
 
 import (
+	"fmt"
+
 	"go.opentelemetry.io/collector/component"
 )
 
 // Config defines configuration for logging exporter.
 type Config struct {
+	// Endpoint defines where we expose our frontend.
```
jmorrell, line 11:
```suggestion
	// Endpoint defines where we serve our frontend app
```

jmorrell (approved):
DALL·E 2023-02-22 15 09 41 - A corgi on a sailboat in port, digital art

main.go
CtrlSpice: Oh oops, this is an artefact of me playing with it. I will put it back.

desktop-exporter/config.go
CtrlSpice: Thank you, that's much better. 😅

PR #102, opened 2023-03-02
Update CI to include integration test

CtrlSpice:
This test runs a bash script that 
- launches the executable in the background 
- sends a minimal test trace using a curl command
- queries the trace store via api call to make sure the trace data was received and processed correctly
- kills the running process and exits with the appropriate exit code

Note: I have added a "CI" environment variable to prevent the browser window from being launched when running in CI. This is set in `go.yaml` under the Integration step.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,10 +12,6 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        include:
-          - os: windows-latest
-            env: 
-              NO_WINDOWS_SERVICE: 1
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -29,3 +29,4 @@ jobs:
 
     - name: Integration
       run: ./.github/workflows/integration.sh
+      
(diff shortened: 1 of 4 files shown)


.github/workflows/integration.sh
```diff
@@ -0,0 +1,31 @@
...
+if [ $rootServiceName == '"test-with-curl"' ]
+then
+    echo 'Exit status 0: All good.'
+    kill -15 $pid
+    exit 0
+else
+    echo 'Exit status 1: unexpected rootServiceName ' + $rootServiceName
+    kill -15 $pid
+    exit 1
+fi
```
jmorrell, line 31: As far as bash scripts go, this is really clean and nice 💯

jmorrell (approved):
DALL·E 2023-03-02 13 22 54 - a corgi sad about debugging a program, digital art

.github/workflows/integration.sh
CtrlSpice: Really?! I guess as far as bash scripts go... 🤣

.github/workflows/go.yml
```diff
@@ -8,7 +8,11 @@ on: push
 jobs:
 
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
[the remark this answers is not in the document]
CtrlSpice: And they pass, too! 😁

PR #103, opened 2023-03-27
Add per platform executables

CtrlSpice:
Adding the `goreleaser.yaml` file to build binaries for MacOS, Linux, and Windows. Yes, we are finally doing this.


jmorrell (approved):
Yay binaries!
DALL·E 2023-03-27 11 56 21 - Two corgis dancing in the moonlight, digital art

PR #114, opened 2023-05-07
Add keyboard navigation for traces

CtrlSpice:
First step in adding keyboard navigation to the UI. You can now go up (arrowLeft, h) and down (arrowRight, l) the trace summaries in the sidebar. 

It's a bit hard to capture, but notice the position of the cursor in the gif:
KeyboardTraces

### DIFF
--- a/desktopexporter/app/components/sidebar-view/trace-list.tsx
+++ b/desktopexporter/app/components/sidebar-view/trace-list.tsx
@@ -161,7 +161,7 @@ export function TraceList(props: TraceListProps) {
     );
   } else {
     selectedTraceID = traceSummaries[selectedIndex].traceID;
-    navigate(`/traces/${selectedTraceID}`);
+    window.location.href = `/traces/${selectedTraceID}`;
   }
 
   // Set up keyboard navigation
--- a/desktopexporter/static/main.js
+++ b/desktopexporter/static/main.js
@@ -51216,7 +51216,7 @@ https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_liter
       );
     } else {
       selectedTraceID = traceSummaries[selectedIndex].traceID;
-      navigate(`/traces/${selectedTraceID}`);
+      window.location.href = `/traces/${selectedTraceID}`;
     }
     let arrowLeftPressed = useKeyPress("ArrowLeft");
     let arrowRightPressed = useKeyPress("ArrowRight");


EddieEldridge:
Looks great 👍

desktopexporter/app/components/sidebar-view/trace-list.tsx
```diff
@@ -144,18 +145,50 @@ type TraceListProps = {
...
-    selectedTraceID = traceSummaries[0].traceID;
+    selectedTraceID = traceSummaries[selectedIndex].traceID;
     window.location.href = `/traces/${selectedTraceID}`;
   }
 
+  // Set up keyboard navigation
+  let arrowLeftPressed = useKeyPress("ArrowLeft");
+  let arrowRightPressed = useKeyPress("ArrowRight");
+  let hPressed = useKeyPress("h");
+  let lPressed = useKeyPress("l");
```
jmorrell, line 171: As a hygiene thing, it would be good if the specific key didn't leak into the variable names. If we want to change them later, or support multiple layouts (not only other language layouts, but dvorak programmers a a thing) then we'll want this to be configurable.

So one more layer of indirection here would be my suggestion. Maybe a `keyMapping` type.

```typescript
type KeyMapping = {
  nextSpan: string;
  prevSpan: string;
  nextTrace: string;
  prevTrace: string;
  // support a second mapping, or maybe use a string[] as a value instead?
  nextSpanAlt: ?string;
  prevSpanAlt: ?string;
  nextTraceAlt: ?string;
  prevTraceAlt: ?string;
}
```

```typescript
let nextSpanPressed = useKeyPress(keyMapping.nextSpan);
```

You could also modify `useKeyPress` to accept an array of keys directly

```typescript
let prevTracePressed = useKeyPress(["ArrowLeft", "h"])
```

jmorrell (changes requested):

desktopexporter/app/components/sidebar-view/trace-list.tsx
CtrlSpice: Hey, so it was my understanding that using event.key in the `useKeyPress` hook (as opposed to event.keyCode) means this will work fine for Dvorak as is.

For the rest though, I agree with you. I'm going to modify `useKeyPress` to take arrays in a separate PR, so I can clean up navigation for trace summaries in the sidebar and the spans in the waterfall view at the same time. 😁

desktopexporter/app/components/sidebar-view/trace-list.tsx
```diff
@@ -144,18 +145,50 @@ type TraceListProps = {
...
+
+  useEffect(() => {
+    if (arrowRightPressed || lPressed) {
+      selectedIndex =
+        selectedIndex < traceSummaries.length - 1
+          ? selectedIndex + 1
+          : traceSummaries.length - 1;
+      selectedTraceID = traceSummaries[selectedIndex].traceID;
+      navigate(`/traces/${selectedTraceID}`);
+    }
```
jmorrell, line 189: These chunks of code are a little mysterious without a comment or a name. Factoring them out into a named function might make this read a bit better

```typescript

if (arrowLeftPressed || hPressed) {
      navigateToPreviousTrace()
}

if (arrowRightPressed || lPressed) {
  navigateToNextTrace()
}
```

jmorrell (approved):
DALL·E 2023-05-14 11 43 42 - a very confused corgi using a keyboard  Floating text says _I have no idea what I am doing_, digital art

PR #129, opened 2023-06-06
Add host flag

svrnm:
Not a go expert, so maybe the implementation needs to look differently.

I wanted to run otel-desktop-viewer in a docker container but didn't get it working because it only listened on localhost. Of course I could build a custom collector and set the config, but I thought this way it is more convenient.

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,6 +34,7 @@ func runInteractive(buildInfo component.BuildInfo) error {
 
 func newCommand(info component.BuildInfo) *cobra.Command {
 	var httpPortFlag, grpcPortFlag, browserPortFlag int
+	var hostFlag string
 
 	rootCmd := &cobra.Command{
 		Use:          info.Command,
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -45,15 +46,15 @@ receivers:
   otlp:
     protocols:
       http:
-        endpoint: localhost:` + strconv.Itoa(httpPortFlag) + `
+        endpoint: ` + hostFlag + `:` + strconv.Itoa(httpPortFlag) + `
       grpc:
-        endpoint: localhost:` + strconv.Itoa(grpcPortFlag) + `
+        endpoint: ` + hostFlag + `:` + strconv.Itoa(grpcPortFlag) + `
```
jmorrell, line 51: This is starting to get a little unwieldy. We should probably explore using a template language in the future @CtrlSpice like https://pkg.go.dev/text/template

jmorrell (approved):
Nice! This looks great as-is

PR #149, opened 2024-09-26
Replace in-memory store with DuckDB

CtrlSpice:
This PR replaces the in-memory list-based store with duckDB:

- So far, traces are the only signal supported, as I wish to break this up into multiple PRs and my first goal was to restore functionality to how it was before the changes.
- Generating trace summaries has been offloaded onto store queries instead of pulling and processing all traces
- I would like to do the same for building and flattening the span tree eventually, but I think this would also be best as a separate PR so we can look at it in isolation.

a charming capybara dressed in a costume that resembles a duck


desktopexporter/internal/store/store.go
```diff
@@ -1,134 +1,288 @@
...
+	defer s.mut.Unlock()
 
-	// Enqueue, then append, as the enqueue process checks if the traceID is already in the map to keep the trace alive
-	store.enqueueTrace(spanData.TraceID)
-	traceData, traceExists := store.TraceMap[spanData.TraceID]
-	if !traceExists {
-		traceData = telemetry.TraceData{
-			TraceID: spanData.TraceID,
-			Spans:   []telemetry.SpanData{},
+	con, err := s.connector.Connect(ctx)
```
jmorrell, line 51: I wonder if we should store the connection in the `Store` so we don't have reconnect each time? Probably not significant at the scales this will be working at though, especially with an in-memory db

desktopexporter/internal/store/store.go
```diff
@@ -1,134 +1,288 @@
...
-	defer store.mut.Unlock()
+	rows, err := conn.QueryContext(ctx, SELECT_TRACE, traceID)
+	if err == sql.ErrNoRows {
+		return trace, telemetry.ErrTraceIDNotFound
+	} else if err != nil {
+		log.Fatalf("could not retrieve spans: %s", err.Error())
+	}
+
+	for rows.Next() {
+		span := telemetry.SpanData{}
```
jmorrell, line 141: Wow, this is kind of painful, but looks right. Factoring this out into a helper function to get a span from a row result might help the code read easier.

desktopexporter/internal/telemetry/sample.go
```diff
@@ -184,7 +184,8 @@ func fillCurrencyScope(scope pcommon.InstrumentationScope) {
 
 // requests scope data
 func fillRequestsScope(scope pcommon.InstrumentationScope) {
-	scope.SetDroppedAttributesCount(0)
+	scope.Attributes().PutStr("pumpkin", "pie")
```
jmorrell, line 187: 🎃🥧

desktopexporter/internal/telemetry/sample.go
```diff
@@ -184,7 +184,8 @@ func fillCurrencyScope(scope pcommon.InstrumentationScope) {
...
+	scope.SetDroppedAttributesCount(42)
```
jmorrell, line 188: I see you

jmorrell (approved):
Looks great! Writing that deserialize code must have been rough

desktopexporter/internal/store/queries.go
```diff
@@ -0,0 +1,60 @@
...
+		CREATE TABLE IF NOT EXISTS spans 
+		(traceID VARCHAR, 
+		traceState VARCHAR, 
+		spanID VARCHAR, 
+		parentSpanID VARCHAR,
+		name VARCHAR, 
+		kind VARCHAR, 
+		startTime TIMESTAMP_NS, 
+		endTime TIMESTAMP_NS,
+		attributes JSON,
```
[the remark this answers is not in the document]
CtrlSpice: If we map, we have to do some type of type of magic with types to account for `map[string]interface{}` so I want to tackle this as a separate PR once I finish integrating logging and metrics :)

desktopexporter/internal/store/store.go
```diff
@@ -1,134 +1,288 @@
...
 
 	"github.com/CtrlSpice/otel-desktop-viewer/desktopexporter/internal/telemetry"
 )
 
 type Store struct {
-	MaxQueueSize int
-	mut          sync.Mutex
-	TraceQueue   *list.List
-	TraceMap     map[string]telemetry.TraceData
+	mut       sync.Mutex
```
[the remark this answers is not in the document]
CtrlSpice: So, the answer is... kinda? We still need some form of concurrency control. A mutex is the simplest way (for me personally, because of the C/Realtime/embedded background) to deal with potentially competing write operations. The other way is to check if we get a SQLITE_BUSY (i.e. database is locked) error and retry but I'm more comfortable with a mutex and the result is the same. I can am not married to the idea though, and will change it if you think the other way is better.

desktopexporter/internal/store/store.go
CtrlSpice: Oh okay, so here the answer is yes and. 

I should store the connection per best practices ([https://duckdb.org/docs/guides/performance/how_to_tune_workloads.html#best-practices-for-using-connections](https://duckdb.org/docs/guides/performance/how_to_tune_workloads.html#best-practices-for-using-connections)), but also I have a `driver.Conn` for adding spans using the DuckDB appender and a `sql.Conn` and I need to figure out how to resolve that.

desktopexporter/internal/telemetry/sample.go
CtrlSpice: I fixed it in the next pr

desktopexporter/internal/telemetry/sample.go
CtrlSpice: This one also

desktopexporter/internal/store/store.go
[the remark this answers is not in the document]
jmorrell: Nope, sounds good to me!

desktopexporter/internal/telemetry/sample.go
jmorrell: This was by no means a request to fix. I like the 🎃🥧

PR #150, opened 2024-09-30
Bringing testing back (part 1)

CtrlSpice:
This test checks that we correctly convert otlp traces to the structs we use internally and send to the front end. It uses the telemetry data generated by `telemetry.sample.go` as test data. 

TODO: I have not included logs and metrics data in testing yet, as their definitions aren't complete. I'll update this test when I work on their respective definitions.

Note: I don't feel like logging into the capybara generating machine, so have here's our queen Moo Deng
moo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ build-go:
 
 .PHONY: test-go
 test-go:
-	cd desktopcollector; go test ./...
+	cd desktopexporter; go test ./...
 	
 .PHONY: run-go
 run-go:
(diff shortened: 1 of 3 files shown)


jmorrell (approved):
These tests look good, but I'm not sure why we're asserting against our sample data? It does exercise our Span struct, but mostly just asserts that you get back out what you put in.

An interesting extension might be to put this data into the store, and then pull it back out again to exercise the span serialization / deserialization logic from #149

CtrlSpice:
@jmorrell So I know we discussed this face-to-face, but I am leaving this here for Future Mila, who reads these things. Because our sample data is generated as `ptrace`, `pmetric`, `plog`, and `pcommon`, we are testing the aggregate and extract functions in `span.go` & friends to make sure that we are unwrapping those data types properly. I will write a separate test for Store in a follow up PR.

PR #151, opened 2024-10-11
Standard library routing

CtrlSpice:
- Changed `server.go` to use standard routing instead of gorilla/mux (thank you for your service)
- Wrote tests for the handlers and routing
- Fixed small bugs in `store.go` and `queries.go`
- Reintroduced CI testing, and the README because I was in there
- Made the AI copilot generate capybaras doing human things for my amusement again

A vibrant digital illustration featuring a capybara wearing a blue kigurumi resembling a gopher. The capybara looks delightful as it waves goodbye. Its friend, a gorilla, is reciprocating with a warm expression, making the scene absolutely charming.


desktopexporter/internal/server/server.go
```diff
@@ -22,57 +22,102 @@ var assets embed.FS
...
+	return s.server.ListenAndServe()
 }
 
-func writeJSON(writer http.ResponseWriter, data any) {
-	jsonTraceData, err := json.Marshal(data)
+func (s *Server) Close() error {
+	return s.server.Close()
+}
+
+func (s *Server) Handler(serveFromFS bool) http.Handler {
```
jmorrell, line 64: `serveFromFS` should probably also be an option passed into `NewServer`. 

Naively, I would expect the routes to be set up as part of `NewServer` not require a further call

desktopexporter/internal/server/server_test.go
```diff
@@ -0,0 +1,321 @@
...
+		assert.Equal(t, uint32(1), testSummaries.TraceSummaries[0].SpanCount)
+	})
+}
+
+func TestTraceIDHandler(t *testing.T) {
+	s, teardown := setupWithTrace(t)
+	defer teardown(t)
+
+	srv := httptest.NewServer(s.Handler(false))
+	defer srv.Close()
```
jmorrell, line 142: Right now this test is slightly different from the others, but they should probably all be the same. I think you could simplify things by moving these bits into `setupWithTrace(t)`. Maybe have it return both `*Server` and `*httptest.Server`

```go
server, testServer, teardown := setupWithTrace(t)
// also calls `testServer.Close()`
defer teardown(t)
```

desktopexporter/internal/server/server_test.go
```diff
@@ -0,0 +1,321 @@
...
+		s.tracesHandler(rec, req)
+		res := rec.Result()
+		defer res.Body.Close()
+
+		b, err := io.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("could not read response: %v", err)
+		}
+
+		if res.StatusCode != http.StatusOK {
```
jmorrell, line 88: Not required, but the `assert` library can be used for these equality assertions:

```go
assert.Equal(t, http.StatusOK, res.StatusCode)
```

desktopexporter/internal/server/server_test.go
```diff
@@ -0,0 +1,321 @@
...
+			t.Errorf("expected status OK; got %v", res.Status)
+		}
+
+		err = json.Unmarshal(b, &testSummaries)
+		if err != nil {
```
jmorrell, line 93: Similarly:

```go
assert.Nil(t, err, "could not unmarshal bytes to trace summaries")
```

though your custom error messages here are a bit more descriptive

https://github.com/stretchr/testify/?tab=readme-ov-file#assert-package

desktopexporter/internal/server/server.go
```diff
@@ -1,16 +1,16 @@
...
 	"encoding/json"
-	"fmt"
 	"io/fs"
 	"log"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
-	"github.com/gorilla/mux"
```
jmorrell, line 13: Unless I'm messing something up locally, this directory needs a `go mod tidy` since we dropped a dependency

jmorrell (approved):
Some suggestions for refinements in this or further PRs, but nothing that should block this from merging

.github/workflow/go.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow will build a golang project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
+
+name: Go
```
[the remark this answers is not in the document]
CtrlSpice: Oh whoops. Yeah, that makes sense.

desktopexporter/internal/server/server.go
```diff
@@ -22,57 +22,102 @@ var assets embed.FS
...
-		writer.WriteHeader(http.StatusOK)
+func NewServer(endpoint string) *Server {
+	s := Server{
+		server: http.Server{
+			Addr: endpoint,
+		},
+		Store: store.NewStore(context.Background()),
 	}
+
+	serveFromFS, err := strconv.ParseBool(os.Getenv("SERVE_FROM_FS"))
```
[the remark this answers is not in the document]
CtrlSpice: I'll do this neatly in its own PR so I don't have to look for it later

desktopexporter/internal/server/server.go
```diff
@@ -22,57 +22,102 @@ var assets embed.FS
...
-func (s *Server) sampleDataHandler() func(http.ResponseWriter, *http.Request) {
-	return func(writer http.ResponseWriter, request *http.Request) {
-		sample := telemetry.NewSampleTelemetry()
-		s.store.AddSpans(request.Context(), sample.Spans)
+func (s *Server) Start() error {
+	defer s.Store.Close()
 
-		//TODO: Add sample logs and metrics
-		writer.WriteHeader(http.StatusOK)
+	_, isCI := os.LookupEnv("CI")
```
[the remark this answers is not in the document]
CtrlSpice: Ditto this.

desktopexporter/internal/server/server.go
```diff
@@ -22,57 +22,102 @@ var assets embed.FS
...
+	return s.server.Close()
+}
+
+func (s *Server) Handler(serveFromFS bool) http.Handler {
+	router := http.NewServeMux()
+	router.HandleFunc("/api/traces", s.tracesHandler)
+	router.HandleFunc("/api/traces/{id}", s.traceIDHandler)
+	router.HandleFunc("/api/sampleData", s.sampleDataHandler)
+	router.HandleFunc("/api/clearData", s.clearTracesHandler)
+	router.HandleFunc("/traces/{id}", indexHandler)
```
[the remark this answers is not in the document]
CtrlSpice: It was my understanding that GET was the default if no verb is specified, and they are all GET, so would that not be redundant? I'm happy to do it, just wanted to clarify.

desktopexporter/internal/server/server.go
CtrlSpice: I'm not sure where else to handle this one, since serveFromFS is only used during development. I'll figure it out in the same PR as the others though.

desktopexporter/internal/server/server_test.go
CtrlSpice: Good call

desktopexporter/internal/server/server_test.go
CtrlSpice: So, super fun thing that I'm going to be laughing about all day but is useful here: `assert.Nilf` is a thing: https://pkg.go.dev/github.com/stretchr/testify@v1.9.0/assert#Nilf

PR #155, opened 2024-11-13
add -db flag for persistent data

banksean:
This hack adds a `--db <path>` flag so that the duckdb instance is persisted to disk between otel-desktop-viewer runs instead of disappearing forever when the process exits.

```
make run-go-db
```

Will create /tmp/otel.db if it doesn't exist, and load its contents if it does.  Traces that show up in the viewer should be persisted to that file.

After stopping the viewer (maybe before that even, thought it's probably not safe), you should be able to query the trace data in that file using the duckdb cli:

```
> duckdb /tmp/otel.db                                                                                                            
v0.10.1 4a89d97db8
Enter ".help" for usage hints.
D .schema
CREATE TABLE spans(traceID VARCHAR, traceState VARCHAR, spanID VARCHAR, parentSpanID VARCHAR, "name" VARCHAR, kind VARCHAR, startTime TIMESTAMP_NS, endTime TIMESTAMP_NS, attributes JSON, events JSON, links JSON, resourceAttributes JSON, resourceDroppedAttributesCount UINTEGER, scopeName VARCHAR, scopeVersion VARCHAR, scopeAttributes JSON, scopeDroppedAttributesCount UINTEGER, droppedAttributesCount UINTEGER, droppedEventsCount UINTEGER, droppedLinksCount UINTEGER, statusCode VARCHAR, statusMessage VARCHAR);
D select count(*) from spans;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│           50 │
└──────────────┘
D 
```

The implementation for this change is admittedly a mess -I don't quite understand the generated code bits work in this repo, so I imagine these changes would probably get clobbered by the next `go generate` run anyways.

Note: this could probably have been more easily implemented via environment variables, and have store.go read those directly instead of trying to pass cli flags through all the layers of config and factory indirection.

### DIFF
--- a/desktopexporter/factory.go
+++ b/desktopexporter/factory.go
@@ -17,10 +17,10 @@ const (
 )
 
 // Creates a factory for the Desktop Exporter
-func NewFactory() exporter.Factory {
+func NewFactory(dbFilename string) exporter.Factory {
 	return exporter.NewFactory(
 		metadata.Type,
-		createDefaultConfig,
+		createDefaultConfig(dbFilename),
 		exporter.WithTraces(createTracesExporter, metadata.TracesStability),
 		exporter.WithMetrics(createMetricsExporter, metadata.MetricsStability),
 		exporter.WithLogs(createLogsExporter, metadata.LogsStability),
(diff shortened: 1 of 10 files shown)


desktopexporter/factory.go
```diff
@@ -17,20 +17,23 @@ const (
 )
 
 // Creates a factory for the Desktop Exporter
-func NewFactory() exporter.Factory {
+func NewFactory(dbFilename string) exporter.Factory {
```
nopcoder, line 20: Not qualified to do a full review here. but I think there is no need to add dbFilename as argument here and pass it to createDefaultConfig.
If you add a line in main.go as part of `set.ConfigProviderSettings.ResolverSettings.URIs` you can set a the value for the db filename using `yaml:exporters::desktop::dbfilename: ` + dbFilenameFlag

Can scope the flag variable to the `newCommand` and have mapstructure populate the value based on the flag.

This way you can skip the value passing and have the value populated based on the flag.

PR #157, opened 2024-11-27
Persistence Support for DuckDB

### LINKED ISSUE #145: Export all the traces to file
The viewer is really handy I love it.

One of the massive perks of having all your traces go to a vendor is the deeper analysis tools that they give you. I am not suggesting this project implements that, but it would be handy to have a simple way of dumping all the traces to a file so that you can analyse the data in other tools. 

Admittedly I could maybe have the collector, or the app for that matter, dump the traces to a file, but I am lazy, and I want a button in a tool. I am running this during benchmarking and property-based testing with generated test input and so I have many traces for the same activity I want to grok.

CtrlSpice:
- Added the `--db` command line flag, that allows used to provide a file path for a persistent DuckDB database. This flag is optional -- omitting it will launch an DuckDB in-memory, same as before.
- The database file path is passed to the `desktopexporter` via config file, using the same mechanism as the browser `endpoint` 
- Wrote unit test that creates and populates a persistent database, closes the connection to the `Store` object, then reopens a new `Store` with the same file path, and retrieves traces to ensure that data is written to disk correctly.
- Closes #145

A charming and fluffy capybara carefully selecting from an array of different flags displayed in a flag store

### DIFF
--- a/desktopexporter/internal/store/store_test.go
+++ b/desktopexporter/internal/store/store_test.go
@@ -1,0 +1,49 @@
+package store
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/CtrlSpice/otel-desktop-viewer/desktopexporter/internal/telemetry"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPersistence(t *testing.T) {
+	ctx := context.Background()
+	store := NewStore(ctx, "./quack.db")
+
+	// Check that db file is created properly
+	_, err := os.Stat("./quack.db")
+	assert.NoErrorf(t, err, "database file does not exist: %v", err)
+
+	// Add sample spans to the store
+	err = store.AddSpans(ctx, telemetry.NewSampleTelemetry().Spans)
+	assert.NoErrorf(t, err, "could not add spans to the database: %v", err)
+
+	// Get trace summaries and check length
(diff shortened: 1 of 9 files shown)


desktopexporter/internal/store/store_test.go
```diff
@@ -0,0 +1,49 @@
...
+	trace, err := store.GetTrace(ctx, "42957c7c2fca940a0d32a0cdd38c06a4")
+	if assert.NoErrorf(t, err, "could not get trace: %v", err) {
+		assert.Equal(t, "37fd1349bf83d330", trace.Spans[0].SpanID)
+	}
+
+	// Clean up
+	err = store.Close()
+	assert.NoErrorf(t, err, "could not close database: %v", err)
+
+	err = os.Remove("./quack.db")
```
jmorrell, line 47: Do you also get a `quack.wal` file here?

jmorrell (approved):
Can you illustrate a capybara in a pile of ducklings? The capybara is very cute. The ducklings are all very cute. Everyone is happy about the situation.

It's cool that just threading down a path is all this took!

desktopexporter/internal/store/store_test.go
CtrlSpice: Yeah but it cleans up after itself on close

PR #158, opened 2024-11-28
DEV and CI Environment Variable Cleanup (Part Deux: the Great Deflagging)

CtrlSpice:
- Setting the `DEV` environment variable to `true` launches the app in dev mode, which avoids recompiling the back-end during front-end development by serving the latter from the filesystem instead of embedded assets. This replaces `SERVE_FROM_FS`, in case I want to cram in more useful dev things. 
-Setting the `CI` environment variable to true starts the back-end without launching the browser, which is useful for running tests in CI.
- These variables are checked when we are starting the `desktopexporter` and passed to the server via the `EnvConfig` struct.
- Server tests and makefile have been updated to reflect the new flow

Illustration of a non-binary individual with short brown hair and glasses. They are asleep at their home office desk. A laptop is open next to them. Next to the laptop, a capybara is also asleep, curled up comfortably on the desk.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,10 @@ test-go:
 run-go:
 	SERVE_FROM_FS=true cd desktopcollector; go run ./...
 
+.PHONY: run-db-go
+run-db-go:
+	SERVE_FROM_FS=true cd desktopcollector; go run ./... --db ../duck.db
+
 .PHONY: build-js
 build-js:
 	cd desktopexporter/internal/app; npx esbuild --bundle main.tsx main.css --outdir=../server/static
--- a/Makefile
+++ b/Makefile
@@ -8,15 +8,15 @@ build-go:
 
 .PHONY: test-go
 test-go:
-	cd desktopexporter; go test ./...
+	cd desktopexporter; DEV=false go test ./...
 	
 .PHONY: run-go
 run-go:
-	cd desktopcollector; go run ./... --dev
+	cd desktopcollector; DEV=true STATIC_DIR="/desktopexporter/internal/server/static/" go run ./...
 
 .PHONY: run-db-go
 run-db-go:
-	cd desktopcollector; go run ./... --dev --db ../duck.db
+	cd desktopcollector; DEV=true STATIC_DIR="/desktopexporter/internal/server/static/" go run ./... --db ../duck.db
 
 .PHONY: build-js
 build-js:
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,11 @@ test-go:
 	
 .PHONY: run-go
 run-go:
-	SERVE_FROM_FS=true cd desktopcollector; go run ./...
+	cd desktopcollector; go run ./... --dev
 
 .PHONY: run-db-go
 run-db-go:
-	SERVE_FROM_FS=true cd desktopcollector; go run ./... --db ../duck.db
+	cd desktopcollector; go run ./... --dev --db ../duck.db
 
 .PHONY: build-js
 build-js:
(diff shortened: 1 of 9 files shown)


desktopexporter/internal/server/server.go
```diff
@@ -122,8 +119,8 @@ func (s *Server) traceIDHandler(writer http.ResponseWriter, request *http.Reques
 	}
 }
 
-func indexHandler(writer http.ResponseWriter, request *http.Request) {
-	if os.Getenv("SERVE_FROM_FS") == "true" {
+func (s *Server) indexHandler(writer http.ResponseWriter, request *http.Request) {
```
jmorrell, line 122: I see this little fix. Yay consistency!

desktopexporter/config.go
```diff
@@ -9,8 +9,12 @@ type Config struct {
 	// Endpoint defines the host and port where we serve our frontend app
 	Endpoint string `mapstructure:"endpoint"`
 
-	// Endpoint defines the path of your database file. Setting an enpty string opens DuckDB in in-memory mode
+	// DbPath defines the path of your database file. Setting an enpty string opens DuckDB in in-memory mode
```
jmorrell, line 12:
```suggestion
	// DbPath defines the path of your database file. Setting an empty string opens DuckDB in in-memory mode
```

desktopexporter/exporter.go
```diff
@@ -20,7 +20,7 @@ type desktopExporter struct {
 }
 
 func newDesktopExporter(cfg *Config) *desktopExporter {
-	server := server.NewServer(cfg.Endpoint, cfg.DbPath)
+	server := server.NewServer(cfg.Endpoint, cfg.DbPath, cfg.IsDev)
```
jmorrell, line 23: We're accumulating options here. I think 3 is fine as an arguments list, but if this grows more we should probably define a config object

desktopexporter/internal/server/server.go
```diff
@@ -23,29 +22,25 @@ var assets embed.FS
 type Server struct {
 	server http.Server
 	Store  *store.Store
+	isDev  bool
 }
 
-func NewServer(endpoint string, dbPath string) *Server {
+func NewServer(endpoint string, dbPath string, isDev bool) *Server {
```
jmorrell, line 28: I don't think we're passing `isDev` to the created `Server` object? In that case it will default to `false`

Makefile
```diff
@@ -12,11 +12,11 @@ test-go:
 	
 .PHONY: run-go
 run-go:
-	SERVE_FROM_FS=true cd desktopcollector; go run ./...
+	cd desktopcollector; go run ./... --dev
```
jmorrell, line 15: Can you double-check that this is working? My read of the code is that a `true` value doesn't get passed to the `Server`

jmorrell (changes requested):
I'm not actually sure this should be a command line option, since that leaks your development process into the interface an end-user is using. I do like passing it down as config vs checking an env var in the middle of the server though.

Can you render Illustration of a non-binary individual with short brown hair and glasses. They are asleep at their home office desk. A laptop is open next to them. On their head a capybara is also asleep, curled up comfortably. Everyone is very, very sleepy and cozy

desktopexporter/config.go
CtrlSpice: I took these out of config so the problem solved itself. I opted to read them when we start the exporter, rather than the collector.

desktopexporter/internal/server/server.go
CtrlSpice: Busted! The fix was actually to access a server variable, but yeah it works for consistency too.

desktopexporter/internal/server/server.go
CtrlSpice: I was not. I am now. 😅

desktopexporter/exporter.go
CtrlSpice: I made `EnvConfig` struct. It holds the environment settings and is just happy to help.

PR #166, opened 2025-02-10
Fix(store) Optimize GetTraceSummaries

CtrlSpice:
# Improved Handling of Traces Without Root Spans

This PR improves how we handle traces that don't have root spans, while also optimizing our data structures for better performance.

## Backend Changes
- Fixed trace summary query to properly return all traces, including those without root spans
- Improved ordering using COALESCE for NULL timestamps, ensuring root spans appear first
- Added clear documentation explaining the query logic
- Added comprehensive tests for traces with and without root spans

## Frontend Changes
- Changed from array to Map for trace summaries (`Map<string, TraceSummaryWithUIData>`)
- Updated types to properly handle optional root spans
- Modified TraceList component to work with Map-based structure
- Maintained existing UI/UX while improving underlying performance

## Technical Details
- Uses Map for O(1) trace lookups instead of array traversal
- Query now handles NULL timestamps with COALESCE for proper ordering
- Test coverage includes mixed cases (traces with/without root spans)

## Testing
- Added new test cases specifically for traces without root spans
- Verified proper handling of mixed trace scenarios
- Confirmed frontend correctly displays both types of traces

### DIFF
--- a/desktopexporter/internal/app/components/sidebar-view/sidebar.tsx
+++ b/desktopexporter/internal/app/components/sidebar-view/sidebar.tsx
@@ -11,14 +11,14 @@ const sidebarCollapsedWidth = 70;
 type SidebarProps = {
   isFullWidth: boolean;
   toggleSidebarWidth: () => void;
-  traceSummaries: TraceSummaryWithUIData[];
+  traceSummaries: Map<string, TraceSummaryWithUIData>;
   numNewTraces: number;
 };
 
 export function Sidebar(props: SidebarProps) {
   let sidebarColour = useColorModeValue("gray.50", "gray.700");
   let { isFullWidth, toggleSidebarWidth, traceSummaries, numNewTraces } = props;
-  let isFullWidthDisabled = traceSummaries.length === 0;
+  let isFullWidthDisabled = traceSummaries.size === 0;
 
   if (isFullWidth) {
     return (
--- a/desktopexporter/internal/app/components/sidebar-view/trace-list.tsx
+++ b/desktopexporter/internal/app/components/sidebar-view/trace-list.tsx
@@ -15,6 +15,8 @@ import { useSize } from "@chakra-ui/react-use-size";
 import { TraceSummaryWithUIData } from "../../types/ui-types";
 import { useKeyCombo, useKeyPress } from "../../utils/use-key-press";
 import { KeyboardHelp } from "../modals/keyboard-help";
+import { getDurationString } from "../../utils/duration";
+import { getDurationNs } from "../../utils/duration";
 
 const sidebarSummaryHeight = 120;
 const dividerHeight = 1;
@@ -41,7 +43,6 @@ function SidebarRow({ index, style, data }: SidebarRowProps) {
   let isSelected = selectedTraceID === traceID;
   let backgroundColour = isSelected ? selectedColor : "";
 
-  console.log(traceSummary);
   if (traceSummary.root) {
     // Add zero-width space after forward slashes, dashes, and dots
     // to indicate line breaking opportunity
@@ -55,6 +56,10 @@ function SidebarRow({ index, style, data }: SidebarRowProps) {
       .replaceAll("-", "-\u200B")
       .replaceAll(".", ".\u200B");
 
+    let rootDuration = getDurationNs(
+      traceSummary.root.startTime,
+      traceSummary.root.endTime
+    );
     return (
       <div style={style}>
         <Divider
@@ -85,7 +90,7 @@ function SidebarRow({ index, style, data }: SidebarRowProps) {
           </Text>
           <Text fontSize="xs">
             {"Root Duration: "}
-            <strong>{traceSummary.root.durationString}</strong>
+            <strong>{getDurationString(rootDuration)}</strong>
           </Text>
           <Text fontSize="xs">
             {"Number of Spans: "}
--- a/desktopexporter/internal/app/components/sidebar-view/trace-list.tsx
+++ b/desktopexporter/internal/app/components/sidebar-view/trace-list.tsx
@@ -21,7 +21,7 @@ const dividerHeight = 1;
 
 type SidebarRowData = {
   selectedTraceID: string;
-  traceSummaries: TraceSummaryWithUIData[];
+  traceSummaries: Map<string, TraceSummaryWithUIData>;
 };
 
 type SidebarRowProps = {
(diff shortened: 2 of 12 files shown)


desktopexporter/internal/app/types/api-types.ts
```diff
@@ -1,11 +1,15 @@
...
+};
+
 export type TraceSummary = {
+  traceID: string;
   hasRootSpan: boolean;
-  rootServiceName: string;
-  rootName: string;
-  rootStartTime: string;
-  rootEndTime: string;
+  rootSpan: RootSpan;
```
jmorrell, line 11: `hasRootSpan` indicates that `rootSpan` is optional, but the type doesn't. Would be:

```typescript
rootSpan?: RootSpan;
```
or
```typescript
rootSpan: RootSpan | null
```

We also likely don't need `hasRootSpan` and can use the presence / lack of presence of `rootSpan` for that signal.

desktopexporter/internal/store/queries.go
```diff
@@ -62,12 +62,49 @@ const (
...
-		ORDER BY MAX(startTime) DESC
+	CREATE_ROOT_SPAN_MACRO = `
+        CREATE MACRO get_root_span_field(trace_id, field) AS (
+            SELECT field
+            FROM spans
+            WHERE traceID = trace_id
+            AND parentSpanID = ''
+            LIMIT 1
+        );
+    `
```
jmorrell, line 73: I think this will execute a full table scan every time you run this macro (unless the query optimizer does some magic?). I think you probably want to use a subquery here instead like you are doing for `spanCount`

https://mode.com/sql-tutorial/sql-sub-queries
https://duckdb.org/docs/sql/expressions/subqueries.html

jmorrell (changes requested):

desktopexporter/internal/app/types/api-types.ts
CtrlSpice: Yeah, I was debating that. This is in part what I meant by “I don’t like how my front end is organized”. I’m going to go change it.

desktopexporter/internal/store/queries.go
CtrlSpice: Yeah, this is correct. I simplified it.

desktopexporter/internal/store/queries.go
```diff
@@ -62,12 +62,25 @@ const (
...
-		SELECT traceID 
-		FROM spans
-		GROUP BY traceID
-		ORDER BY MAX(startTime) DESC
-	`
+	// SELECT_TRACE_SUMMARIES retrieves all traces, with root spans first.
+	// Uses COALESCE(startTime, far_future_date) in ORDER BY to handle NULL timestamps:
+	// - For traces with root spans: uses actual startTime
+	// - For traces without root spans: uses far future date (9999-12-31)
+	// This ensures traces without root spans appear after those with root spans in DESC order
```
jmorrell, line 69: I'm not sure this is desired behavior. No root span likely means one of the following is true:

- the trace is still running and not finished yet
- something likely went wrong and the program crashed, especially during development. We likely don't want to hide that trace at the bottom.
- config error, the trace began elsewhere and isn't being forwarded to this service for some reason

Can we use the earliest timestamp in the trace maybe? 🤔

desktopexporter/internal/store/queries.go
```diff
@@ -62,12 +62,25 @@ const (
...
+            CASE WHEN s.parentSpanID = '' THEN s.name END as root_name,
+            CASE WHEN s.parentSpanID = '' THEN s.startTime END as start_time,
+            CASE WHEN s.parentSpanID = '' THEN s.endTime END as end_time,
+            COUNT(*) OVER (PARTITION BY s.traceID) as span_count
+        FROM spans s
+        ORDER BY 
+            s.traceID,
+            s.parentSpanID = '' DESC,
+            COALESCE(s.startTime, '9999-12-31'::timestamp) DESC
+    `
```
jmorrell, line 83: So much nicer!

desktopexporter/internal/app/types/ui-types.ts
```diff
@@ -21,24 +21,20 @@ export type SpanWithUIData =
...
-  | {
-      hasRootSpan: false;
-      spanCount: number;
-      traceID: string;
-    };
+export type RootSpanWithUIData = {
+  serviceName: string;
+  name: string;
+  durationString: string;
+};
```
jmorrell, line 28: Consider dropping this. This is no longer significantly different that `RootSpan` and we can likely just calculate the duration in-line in the render

desktopexporter/internal/app/components/sidebar-view/sidebar.tsx
```diff
@@ -11,14 +11,14 @@ const sidebarCollapsedWidth = 70;
 type SidebarProps = {
   isFullWidth: boolean;
   toggleSidebarWidth: () => void;
-  traceSummaries: TraceSummaryWithUIData[];
+  traceSummaries: Map<string, TraceSummaryWithUIData>;
```
jmorrell, line 14: I know `Map` preserves insertion order, but it feels wrong somehow to rely on this behavior vs a sorted array

desktopexporter/internal/app/components/sidebar-view/trace-list.tsx
```diff
@@ -34,22 +34,23 @@ function SidebarRow({ index, style, data }: SidebarRowProps) {
...
-    selectedTraceID && selectedTraceID === traceSummary.traceID ? true : false;
 
+  // Get traceID for this index
+  let traceID = Array.from(traceSummaries.keys())[index];
+  let traceSummary = traceSummaries.get(traceID)!;
+  let isSelected = selectedTraceID === traceID;
   let backgroundColour = isSelected ? selectedColor : "";
 
-  if (traceSummary.hasRootSpan) {
+  console.log(traceSummary);
```
jmorrell, line 44: nit: delete

jmorrell (approved):
Just a few notes. Nothing required

desktopexporter/internal/store/queries.go
CtrlSpice: You are totally right! I didn’t think about how actual usability here with genuinely broken traces here- I only thought about the unfinished ones and was like “Heh, it will pop back into order when the root gets there”. Good call.

 I just fixed it to sort by root span start if we have one, and fall back to easiest span start if we do not. Should be better.

CtrlSpice: Also the robit helped me write tests! 😁

PR #168, opened 2025-02-12
Allow running in Docker

davetron5000:
I run my dev environment using Docker, and don't want to/can't easily install stuff on my computer with Homebrew.

The `Dockerfile` in this PR worked for me, but I want to contribute whatever makes sense to this project if others want to do the same.

I'm not sure the best way, but I can think of a few scenarios:

1. **Publish a Docker image others can use.** This would require a `Dockerfile` similar to what is in this PR as well as necessary
   scripts and credentials to build and push the image.  It would also require this repo's owners to maintain that image.
2. **Just provide instructions.** This would involve adding something to the `README` I guess?

I'm happy to do more work along either option, so if one is prefered over the other, I will have more questions on how to properly
proceed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,60 @@ Running the CLI will open a browser tab to `localhost:8000` to load the UI,
 and spin up a server listening on `localhost:4318` for OTLP http payloads and
 `localhost:4317` for OTLP grpc payloads.
 
+#### via Docker
+
+The included `Dockerfile` will build a relatively slim image you can use to run otel-desktop-viewer without
+using Homebrew or setting up Go.
+
+To build the image
+
+* [Install Docker](https://docs.docker.com/get-started/)
+  - For Windows or Mac, you likely want Docker Desktop
+  - For Linux, Docker Engine is fine, but do read the [post-install instructions](https://docs.docker.com/engine/install/linux-postinstall/) so that you can run Docker as your user
+* Build the image. You'll need a tag so you can reference it.
+
+  ```
+  > docker build --tag otel-desktop-viewer:alpine-3
+  ```
+
+To use the image, you have a few options:
+
+* **Run it so apps on your computer can access it**:
+
+  ```
+  docker run -p 8000:8000 -p 4317:4317 -p 4318:4318 davetron5000/otel-desktop-viewer:alpine-3
+  ```
+
+  The `-p` arguments map otel-desktop-viewers ports to the same ports on localhost.  When you do this, you would
+  access the web UI via `localhost:8000` and use `localhost:4318` or `localhost:4319` to export to.
+* **Run it in docker compose as part of a devcontainer-based environment**. Here, the app you are working on
+that is sending telemetry is running as a Docker container.  You can use Docker compose to allow that app to
+export to your otel-desktop-viewer, but still access its web UI on your computer. A very simplified
+`docker-compose.yml`:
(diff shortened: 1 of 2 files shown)


alejo-rivera:
Love this since it's exactly what I was about to implement myself. Some thoughts/recommendations, feel free to take it or leave it. 

- Could make this a [multi stage build](https://docs.docker.com/build/building/multi-stage/) to reduce the shipped image size. 
  - i.e. Add another `FROM` (using alpine or something otherwise smaller than `golang`) after installing the binary and then copying just the binary to that new stage of the build. 
- Like you noted this could probably be better integrated with the existing repo's build process (which I'm not familiar with). This could remove the need to running `go install` all together in the docker image and could just use the existing build artifacts. 

At any rate thanks for sharing and I'll be using something based on this until it's merged!

davetron5000:
Good idea - I could not get alpine to work, but ubuntu does work - see latest update.

I'm sure alpine could be made to work, but I don't understand Go at all really.

I also think this needs to be multi-arch build, but I could not get that working on Linux.  Will try on macOS in a bit

alejo-rivera:
Quick glance at [docker hub](https://hub.docker.com/_/golang) seems to show that golang uses debian (ubuntu is debian based too) for their default image (the one you're using in your first stage) 

Switching both stages to an alpine based image should "just work". 


The change would be something like this. 
```dockerfile
FROM golang:1.24-alpine AS golang
...
FROM alpine:24.04
...
```

Agreed on the need for multi-arch. That falls our of my expertise though. I think that's also done at build time by just specifying it to the docker build command and publishing properly ¯\_(ツ)_/¯

davetron5000:
Ah, that worked, and very good savings on space

```
> docker image ls
REPOSITORY                         TAG             IMAGE ID       CREATED             SIZE
davetron5000/otel-desktop-viewer   alpine-3        4bdd50155b22   3 minutes ago       45MB
davetron5000/otel-desktop-viewer   ubuntu-2404     40ae7a5c7925   About an hour ago   115MB
davetron5000/otel-desktop-viewer   go-1.24         34d740f497eb   About an hour ago   1.5GB
```

davetron5000:
It's on docker hub at `davetron5000/otel-desktop-viewer:alpine-3` if you want to try it

alejo-rivera:
> It's on docker hub at `davetron5000/otel-desktop-viewer:alpine-3` if you want to try it

Works like a charm, thanks!

davetron5000:
If we want to merge this, what do we need to do?  Should I add a blurb in the README on how to build and use this?  Without a multi-arch build, that's probably the next best way for someone to use this with Docker

alejo-rivera:
Sounds about right to me. Though I definitely hold no power here and merging is a question for @CtrlSpice.

README.md
```diff
@@ -42,6 +42,60 @@ Running the CLI will open a browser tab to `localhost:8000` to load the UI,
 and spin up a server listening on `localhost:4318` for OTLP http payloads and
 `localhost:4317` for OTLP grpc payloads.
 
+#### via Docker
```
davetron5000, line 45: Documented how to use this.  Tried to not overexplain docker but also not leave the reader completely helpless.

CtrlSpice:
Hey there! Thank you very much for this. Merging it now!

PR #175, opened 2025-05-27
Union Strong 

CtrlSpice:
## go-duckdb supports unions now, and so do we
- Replaced the json-based workaround in we were using in `store.go` to save attributes with proper duckdb.Map and duckdb.Union support.
- Replaced the other json-based workaround we were using to save structs (links and events). This took a bit of manual conversion in `util.go` because of the nested duckdb types, but it is still much better than marshalling and unmarshalling the whole thing.
- A good chunk of util.go (`toDuckDBMap()`, `getListTypeTag()`, `validateUniformList()`, and `stringifyOnOverflow()`) handle type safety things to make sure our attributes follow [OTel specs](https://opentelemetry.io/docs/specs/otel/common/#attribute,) and do not lose accuracy by being stored as the wrong type in our Union. I also did my best to fall back gracefully and warn the user if they do something egregiously outside those specs; I don't think anyone will actually try to cram a list of `unit64`s into an attribute, but in case they do, they will get warned, and the list will fall back to []string. I plan to switch to structured logs in a follow-up PR so I can emit proper warnings because this is getting long. 
- Tested all the things. Added tests for the resource and scope attributes, as mentioned in the last PR.

That's about it, really.
capyunion


desktopexporter/internal/store/queries.go
```diff
@@ -87,8 +87,8 @@ const (
 			droppedLinksCount,
 			statusCode, 
 			statusMessage
-		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?::JSON::MAP(VARCHAR, attribute), ?::JSON::event[], ?::JSON::link[], 
-		 ?::JSON::MAP(VARCHAR, attribute), ?, ?, ?, ?::JSON::MAP(VARCHAR, attribute), ?, ?, ?, ?, ?, ?)
```
jmorrell-cloudflare, line 91: 👋 Bye JSON hack

desktopexporter/internal/store/queries.go
```diff
@@ -87,8 +87,8 @@ const (
 			droppedLinksCount,
 			statusCode, 
 			statusMessage
-		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?::JSON::MAP(VARCHAR, attribute), ?::JSON::event[], ?::JSON::link[], 
-		 ?::JSON::MAP(VARCHAR, attribute), ?, ?, ?, ?::JSON::MAP(VARCHAR, attribute), ?, ?, ?, ?, ?, ?)	 
+		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?::MAP(VARCHAR, attribute), ?::event[], ?::link[],
```
jmorrell-cloudflare: Since we're inserting a `duckdb.Map` do we still need the `?::MAP(VARCHAR, attribute)` conversion? Same with the event and link conversions.

desktopexporter/internal/store/queries.go
```diff
@@ -119,8 +119,7 @@ const (
             s.traceID
     `
 
-	// DuckDB's Go bindings have limited support for UNIONs
-	// So we need to cast the attributes to JSON and then parse them back into the original type
+	// DuckDB's Go bindings now have proper support for UNIONs 🎉
```
jmorrell-cloudflare, line 94: Nice to celebrate, but this doesn't need to be in the code

desktopexporter/internal/store/store.go
```diff
@@ -73,39 +72,34 @@ func (s *Store) AddSpans(ctx context.Context, spans []telemetry.SpanData) error
...
-		resourceAttrs := MarshalAttributes(span.Resource.Attributes)
-		scopeAttrs := MarshalAttributes(span.Scope.Attributes)
-		events := MarshalEvents(span.Events)
-		links := MarshalLinks(span.Links)
-
-		_, err := stmt.ExecContext(ctx,
+	defer appender.Close()
+
+	for _, span := range spans {		
+		// Pass empty arrays instead of nil
```
jmorrell-cloudflare, line 82: Comment isn't clear. Do we need this? What should a future dev keep in mind here? or should it be a comment on the `toDuckDBEvents` function?

desktopexporter/internal/store/store.go
```diff
@@ -73,39 +72,34 @@ func (s *Store) AddSpans(ctx context.Context, spans []telemetry.SpanData) error
...
+		events := toDuckDBEvents(span.Events)
+		links := toDuckDBLinks(span.Links)
```
jmorrell-cloudflare, line 84: Why not convert `events` and `links` in line like `span.Attributes`?

desktopexporter/internal/store/store.go
```diff
@@ -146,15 +144,14 @@ func (s *Store) GetTrace(ctx context.Context, traceID string) (telemetry.TraceDa
 			},
 		}
 
-		// DuckDB's Go bindings have no support for UNIONs
-		// So we are leveraging duckdb's json functionality to fool it into doing the right thing
+		// DuckDB's Go bindings now have proper support for UNIONs 🎉
```
jmorrell-cloudflare, line 147: Same, delete

desktopexporter/internal/store/store_test.go
```diff
@@ -294,10 +307,315 @@ func TestAttributeTypes(t *testing.T) {
...
+			EndTime:      baseTime.Add(time.Second),
+			Attributes: map[string]any{
+				// []any arrays where all elements are the same type (like OpenTelemetry Go)
+				"any_string_list": []any{"apple", "banana", "cherry"},
+				"any_int_list":    []any{int64(10), int64(20), int64(30)},
+				"any_float_list":  []any{float64(1.5), float64(2.5), float64(3.5)},
+				"any_bool_list":   []any{true, false, true},
+				
+				// Mixed integer types in []any (should all become bigint_list)
+				"any_mixed_ints":  []any{int(1), int32(2), int64(3), uint(4), uint32(5)},
```
jmorrell-cloudflare, line 401: Given that the data here is being deserialized from OTLP either protobuf or JSON, I'm not sure that we'd ever see data like this. You would have to build the data up in Go itself iiuc

desktopexporter/internal/store/store_test.go
```diff
@@ -455,6 +796,117 @@ func TestEventsAndLinks(t *testing.T) {
...
+	linkList2 := link2.Attributes["link_list2"].([]any)
+	assert.Len(t, linkList2, 3)
+	assert.Equal(t, true, linkList2[0])
+	assert.Equal(t, false, linkList2[1])
+	assert.Equal(t, true, linkList2[2])
+	
 	assert.Equal(t, uint32(2), link2.DroppedAttributesCount)
 }
+
+func TestResourceAndScopeAttributes(t *testing.T) {
```
jmorrell-cloudflare, line 810: 💯

desktopexporter/internal/store/util.go
```diff
@@ -1,80 +1,52 @@
...
-		return "[]"
-	}
-
-	return string(jsonLinks)
+// duckDBLink represents a link in DuckDB format.
+// NOTE: The `db` struct tags are absolutely required! They map Go struct fields
+// to the correct DuckDB column names. Without them, the database operations will fail.
+type duckDBLink struct {
+	TraceID                string `db:"traceID"`
+	SpanID				   string `db:"spanID"`
```
jmorrell-cloudflare, line 40: funky indentation. Are you using `go fmt`?

desktopexporter/internal/store/util.go
```diff
@@ -1,80 +1,52 @@
...
-	if err != nil {
-		log.Printf("error marshalling links %v: %v", links, err)
-		return "[]"
-	}
-
-	return string(jsonLinks)
+// duckDBLink represents a link in DuckDB format.
+// NOTE: The `db` struct tags are absolutely required! They map Go struct fields
+// to the correct DuckDB column names. Without them, the database operations will fail.
+type duckDBLink struct {
```
jmorrell-cloudflare, line 38: Inconsistent naming. One of these two should likely change.

CtrlSpice: 🦆💥

desktopexporter/internal/store/store.go
CtrlSpice: good call

desktopexporter/internal/store/store_test.go
CtrlSpice: So, from what I understand, it is very difficult but still possible to mess it up if you are working on something that mutates data. Since this is a local dev tool, I'd rather assume that someone at some point will use it to debug something weird, and just provide the additional safeguard.

desktopexporter/internal/store/util.go
CtrlSpice: Good call. I was missing the formatter tool in my editor settings so it wasn't running on save.

PR #181, opened 2025-06-11
Feat/logs back end

CtrlSpice:
Hi! Hello! This PR enables us to receive logs, store them in duckDB, and get them to our front-end.

I also changed how we handle timestamps because the way I was doing it before was not well thought out and became a problem for falling back to observed timestamp for sorting logs when a record's timestamp was unavailable. Anyway, timestamps are now being stored as [UnixNano](https://pkg.go.dev/time#Time.UnixNano), and is marshalled to JSON like this to avoid losing nanosecond precision:
`
// PreciseTimestamp represents a timestamp split into milliseconds and nanoseconds for JSON serialization
type PreciseTimestamp struct {
	Milliseconds int64 `json:"milliseconds"`
	Nanoseconds  int32 `json:"nanoseconds"`
}
`
I had to thread this functionality through on the front-end in trace summaries, spans, and events as well, and get rid of the code that parsed string-based timestamps, so that accounts for good chunk of the changes.

Finally, I added a route for logs behind a feature flag to print out log json. It's a bit of a hack with our current front end, but it's just to confirm that it works:

isBigIsHeavyIsWood

Gemini_Generated_Image_arcip5arcip5arci


desktopexporter/internal/app/types/precise-timestamp.ts
```diff
@@ -0,0 +1,40 @@
+export class PreciseTimestamp {
+  constructor(
+    public milliseconds: number,
+    public nanoseconds: number
```
jmorrell, line 4: Linking otel-js' HrTime: https://github.com/open-telemetry/opentelemetry-js/blob/f2cfd1327a5b131ea795301b10877291aac4e6f5/api/src/common/Time.ts#L30
https://github.com/open-telemetry/opentelemetry-js/blob/f2cfd1327a5b131ea795301b10877291aac4e6f5/packages/opentelemetry-core/src/common/time.ts

We might also simplify our lives by using a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) internal representation but serializing to the 2 numbers on the wire. otel-js didn't do this due to needing to support legacy Node versions, now no longer an issue: https://github.com/open-telemetry/opentelemetry-js/issues/5680

otel-js is now also moving in this direction: https://github.com/open-telemetry/opentelemetry-js/pull/5522

desktopexporter/internal/app/utils/duration.ts
```diff
@@ -1,76 +1,106 @@
...
+      endTime: endTime,
+      milliseconds: 0,
+      nanoseconds: 0,
+      label: ""
+    };
+  }
 
-  return endTimeNs - startTimeNs;
-}
+  // Calculate duration in milliseconds and nanoseconds separately
```
jmorrell, line 52: This logic likely belongs in the `PreciseTimestap` class, maybe as a class method. It's tied to the internal representation of the class.

desktopexporter/internal/app/utils/array-to-tree.ts
```diff
@@ -84,8 +84,13 @@ export function arrayToTree(spans: SpanData[]): RootTreeItem[] {
...
       let earliestStartTimeB = getEarliestStartTime(lookup[b].children);
-
-      return earliestStartTimeA - earliestStartTimeB;
+      
+      if (earliestStartTimeA.isBefore(earliestStartTimeB)) {
+        return -1;
+      } else if (earliestStartTimeA.isAfter(earliestStartTimeB)) {
+        return 1;
+      }
+      return 0;
```
jmorrell, line 93: A `PreciseTimestamp.cmp` function might clean this up, but I kind of like the explicitness here

desktopexporter/internal/app/utils/feature-flags.ts
```diff
@@ -0,0 +1,12 @@
...
+  LOGS_VIEW: false, // Enable/disable the logs view
+} as const;
+
+// Type for feature flag names
+export type FeatureFlag = keyof typeof FEATURE_FLAGS;
+
+// Helper function to check if a feature is enabled
+export function isFeatureEnabled(flag: FeatureFlag): boolean {
+  return FEATURE_FLAGS[flag];
+}
```
jmorrell, line 12: I don't think this is used anywhere?

desktopexporter/internal/server/server.go
```diff
@@ -51,8 +51,17 @@ func (s *Server) Handler() http.Handler {
...
+	router.HandleFunc("GET /api/clearTraces", s.clearTracesHandler)
 	router.HandleFunc("GET /traces/{id}", s.indexHandler)
+	
+	// Feature flag for logs frontend route
+	if os.Getenv("ENABLE_LOGS") == "true" {
+		router.HandleFunc("GET /logs", s.indexHandler)
+		router.HandleFunc("GET /api/logs", s.logsHandler)
+		router.HandleFunc("GET /api/logs/{id}", s.logIDHandler)
+		router.HandleFunc("GET /api/logs/trace/{id}", s.logsByTraceHandler)
+		router.HandleFunc("GET /api/clearLogs", s.clearLogsHandler)
```
jmorrell, line 63: I'm unsure about this API structure, but let's roll with it for now. I expect this to change as we try to build a UI

desktopexporter/internal/telemetry/logs.go
```diff
@@ -11,18 +14,27 @@ type LogsPayload struct {
...
+	SeverityText           string `json:"severityText,omitempty"`
+	SeverityNumber         int32  `json:"severityNumber,omitempty"`
+	
+	Body                   any            `json:"body,omitempty"`
+	Resource               *ResourceData  `json:"resource"`
+	Scope                  *ScopeData     `json:"scope"`
+	Attributes             map[string]any `json:"attributes,omitempty"`
+	DroppedAttributesCount uint32         `json:"droppedAttributeCount,omitempty"`
+	Flags             	   uint32         `json:"flags,omitempty"`
+	EventName              string         `json:"eventName,omitempty"`
```
jmorrell, line 32: Some weird spacing. This could need a `gofmt`

jmorrell (approved):
an image of a capybara wearing a plaid lumberjack shirt falling off a log it was balancing on into a river. The log was floating in the river along with many other logs. The style is cartoonish and reminiscent of a national park poster. Text at the top reads "The Logs must flow"

desktopexporter/exporter.go
```diff
@@ -36,15 +36,15 @@ func (exporter *desktopExporter) pushTraces(ctx context.Context, traces ptrace.T
...
+		fmt.Println(metricsData)
 	}
 	return nil
 }
 
 func (exporter *desktopExporter) pushLogs(ctx context.Context, logs plog.Logs) error {
 	logDataSlice := telemetry.NewLogsPayload(logs).ExtractLogs()
 	for _, logData := range logDataSlice {
-		fmt.Println(logData.ID())
+		fmt.Println(logData)
```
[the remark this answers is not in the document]
CtrlSpice: Good call! It would also be a good time to call
`exporter.server.Store.AddLogs(ctx, logDataSlice)` instead.

desktopexporter/internal/app/types/precise-timestamp.ts
CtrlSpice: Super good call. I will do this in its own PR for my own mental health, since I will have to re-thread it through all the time/duration stuff, but it will simplify my life significantly.

desktopexporter/internal/app/main.tsx
```diff
@@ -6,33 +6,48 @@ import { extendTheme, type ThemeConfig } from "@chakra-ui/react";
...
   initialColorMode: "system",
-  useSystemColorMode: false,
+  useSystemColorMode: true,
 };
-const theme = extendTheme({ config });
+let theme = extendTheme({ config });
 
-const router = createBrowserRouter([
+// Check feature flag for logs
+let enableLogs = localStorage.getItem('enableLogs') === 'true';
```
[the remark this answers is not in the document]
CtrlSpice: I'll switch it out when I'm implementing the permanent front end. localStorage will do for the super basic prototyping I'm using to make sure the logs (and soon metrics) API is APIing.

desktopexporter/internal/telemetry/logs.go
CtrlSpice: Note mostly to self because I will forget and break my editor settings again in the future: For gofmt to work on save, we need `"editor.formatOnSaveMode": "file"`

desktopexporter/internal/app/types/precise-timestamp.ts
```diff
@@ -0,0 +1,40 @@
+export class PreciseTimestamp {
```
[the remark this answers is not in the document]
CtrlSpice: I'ma get rid of it internally and unmarshal to BigInt

PR #203, opened 2025-09-30
handle out-of-order root span

bendrucker:
Fixes a non-deterministic ordering bug where valid traces would report "missing a parent span." I observed this sending trace data from the [Python client](https://opentelemetry.io/docs/languages/python/), via the [FastAPI instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/fastapi/fastapi.html). 

Screenshot 2025-09-29 at 20 51 14

## Issue

The application uses this in `SelectTraceSummaries`:

```sql
COALESCE(
    MIN(CASE WHEN s.ParentSpanID = '' THEN s.StartTime END) OVER (PARTITION BY s.TraceID),
    MIN(s.StartTime) OVER (PARTITION BY s.TraceID)
) DESC
```

But this has the same value for every span in a given trace ID. And so the `ORDER BY` clause effectively becomes a noop, returning a random span. 

https://github.com/CtrlSpice/otel-desktop-viewer/blob/730f64a63e9a90dd799a20b4b6366d35cc042dc2/desktopexporter/internal/store/queries.go#L215-L226

If this query ends up selecting the wrong span, `CASE WHEN s.ParentSpanID = ''` doesn't match. 

## Fix

Adding this ensures that a root span will always be the first distinct result:

```sql
CASE WHEN s.ParentSpanID = '' THEN 0 ELSE 1 END
```

I ran out of debugging patience to dig into exactly why this client managed to reproduce this issue, e.g., is the root span not sent first? This issue seemed to occur deterministically so I suspect that's the case. But I can confirm the fix works:

Screenshot 2025-09-29 at 21 15 02

### DIFF
--- a/desktopexporter/internal/store/queries.go
+++ b/desktopexporter/internal/store/queries.go
@@ -233,12 +233,13 @@ const (
             CASE WHEN s.ParentSpanID = '' THEN s.EndTime END as end_time,
             COUNT(*) OVER (PARTITION BY s.TraceID) as span_count
         FROM spans s
-        ORDER BY 
+        ORDER BY
             COALESCE(
                 MIN(CASE WHEN s.ParentSpanID = '' THEN s.StartTime END) OVER (PARTITION BY s.TraceID),
                 MIN(s.StartTime) OVER (PARTITION BY s.TraceID)
             ) DESC,
-            s.TraceID
+            s.TraceID,
+            CASE WHEN s.ParentSpanID = '' THEN 0 ELSE 1 END
     `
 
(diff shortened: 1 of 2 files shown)


YOU54F:
Cheers for the fix, running locally and it helped me when trying to test tracing between ruby & rust client apps, and a ruby server app 👍🏾 

Had pulled down the code to take a look but spotted your PR 🙌🏾

CtrlSpice:
@bendrucker Thank you so much for this fix. Seriously. It was driving me bonkers and I could not pinpoint it. You rock!

PR #207, opened 2025-10-23
Fix broken image link

gstelmach-godaddy:


CtrlSpice:
Thanks for the fix! The README has since been rewritten and the broken image link no longer exists, so closing as obsolete.

PR #213, opened 2026-03-23
feat: add CLI flags to override OTLP HTTP receiver URL paths

### LINKED ISSUE #170: feature request: endpoint override
First off, thanks for creating this wonderful tool!

Feature request: ability to override the default endpoint (endpoints?)
* for example, if an app doesn't want to use `/v1/traces` and `/v1/metrics`, allow specifying something different
  * when initializing traces and metrics, each has a `WithUrlPath()` method that overrides the defaults.

If my other feature request is implemented, I think the custom endpoints should also be used there.

RichardoMrMu:
## What
Add three new CLI flags to allow overriding the URL paths used by the OTLP HTTP receiver.

## Why
Closes #170

Some applications use the OTel SDK's `WithURLPath()` client option to send telemetry to non-standard paths instead of the defaults (`/v1/traces`, `/v1/metrics`, `/v1/logs`). Currently there is no way to configure the receiver to listen on those custom paths.

## How
Added three new CLI flags that map directly to the `otlpreceiver`'s HTTP config:

| Flag | Default | Maps to |
|------|---------|---------|
| `--traces-url-path` | `/v1/traces` | `receivers.otlp.protocols.http.traces_url_path` |
| `--metrics-url-path` | `/v1/metrics` | `receivers.otlp.protocols.http.metrics_url_path` |
| `--logs-url-path` | `/v1/logs` | `receivers.otlp.protocols.http.logs_url_path` |

The defaults match the OTel spec, so existing behavior is unchanged.

## Usage
```bash
# Use custom URL paths
otel-desktop-viewer --traces-url-path /custom/traces --metrics-url-path /custom/metrics

# Default behavior unchanged
otel-desktop-viewer
```

## Notes
- Only the HTTP receiver paths are configurable; gRPC uses service reflection and doesn't have URL paths.
- The `otlpreceiver` internally validates and sanitizes the paths (ensures leading `/`, etc.).

### DIFF
--- a/main.go
+++ b/main.go
@@ -57,6 +57,7 @@ func runInteractive(params otelcol.CollectorSettings) error {
 func newCommand(set otelcol.CollectorSettings) *cobra.Command {
 	var httpPortFlag, grpcPortFlag, browserPortFlag int
 	var hostFlag, dbFlag string
+	var tracesURLPathFlag, metricsURLPathFlag, logsURLPathFlag string
 	var openBrowserFlag bool
 
 	rootCmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


CtrlSpice:
Thanks for the contribution! The motivating issue (#170) has since been closed and main.go has changed substantially, so I'm closing this to tidy the queue. If path overrides become a priority again we'd welcome a fresh PR against the current code.

PR #228, opened 2026-07-01
Add --metrics-port flag for internal Prometheus self-telemetry

jschaf:
## Problem

The collector's internal self-telemetry Prometheus endpoint is hardcoded to `:8888`, so `otel-desktop-viewer` fails to start when something already holds that port:

```
Error: failed to create SDK: binding address localhost:8888 for Prometheus exporter: listen tcp 127.0.0.1:8888: bind: address already in use
```

`:8888` is a commonly-occupied port (Jupyter, other collectors, SSH tunnels). The OTLP/UI ports are all configurable via flags, but this one isn't — and `config.go` actively reserves `8888`, so there's no escape hatch. On a machine that already has `:8888` bound, the viewer simply won't launch.

## Change

Add a `--metrics-port` flag (default `8888`, so existing behavior is unchanged) that sets the `service::telemetry::metrics` Prometheus reader explicitly, letting users relocate it:

```bash
otel-desktop-viewer --metrics-port 18888
```

It's a one-line config URI wired to a new flag, matching the existing `--http`/`--grpc`/`--browser-port` pattern, plus a README line.

## Verification

Built from source (Go 1.26, collector v0.126.0) and, with `:8888` already occupied:

- default (no flag) → still errors with the bind message above (unchanged behavior)
- `--metrics-port 18888` → starts cleanly; UI serves on `:8000`, and the internal metrics are served on `:18888` (`GET /metrics` → 200)

Happy to adjust the flag name/semantics (e.g. a `--disable-internal-metrics` variant instead) if you'd prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ Flags:
       --grpc int           OTLP gRPC listen port (default 4317)
       --host string        Host for OTLP receivers and the web UI (default localhost)
       --http int           OTLP HTTP listen port (default 4318)
+      --metrics-port int   Port for the collector's internal Prometheus self-telemetry (default 8888)
       --open-browser       Open the browser on launch (default true)
   -h, --help               help for otel-desktop-viewer
   -v, --version            version for otel-desktop-viewer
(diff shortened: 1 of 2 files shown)


jschaf:
Dupe of https://github.com/CtrlSpice/otel-desktop-viewer/pull/213

PR #229, opened 2026-07-01
Make traces deep-linkable via /traces/{id}

jschaf:
Currently, you cannot deep-link a trace. The selected trace lives in component state (`selectedTraceId` in `TracesPage.svelte`), not the URL.

- You can't share or bookmark a trace, and a refresh drops the selection.
- `/traces/{id}` is handled by the Go server and renders blank.
- a fresh page load of `/traces`, `/metrics`, or `/logs` 404s, since the server only serves the SPA for `/` and `/traces/{id}`.

Modify the server to serve index.html for deep links and let the Svelte router render the appropriate page.

### DIFF
--- a/desktopexporter/internal/frontend/src/App.svelte
+++ b/desktopexporter/internal/frontend/src/App.svelte
@@ -23,7 +23,7 @@
   <Route path="/logs">
     <LogsPage />
   </Route>
-  <Route path="/traces">
+  <Route path="/traces/*">
     <TracesPage />
   </Route>
 </main>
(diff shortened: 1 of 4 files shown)


CtrlSpice:
Hey, thank a bunch for this. 

Funny timing: I've been working on URL state in a parallel branch that makes traces, metrics, and logs deep-linkable, including the sub-views in metrics and the time window in the query string. So the frontend half of this PR is about to be superseded when that branch lands.

I never got to the server code though, so a hard refresh on /metrics or /logs would still 404. Your spaHandler is a much better shape than the /traces/{id} special case it replaces, and the tests are appreciated.

I'm going to merge this as-is and rebase my routing branch on top. Your server changes will stay put, but thanks for digging into both halves of this.

PR #282, opened 2026-07-26
Pin GitHub Actions to commit SHAs

CtrlSpice:
## Summary
- Pins all 31 action references across the three workflows to full commit SHAs, with the version tag preserved as a comment (e.g. `actions/checkout@3d3c42e... # v7`).
- Mutable tags like `@v7` can be moved by whoever controls the action's repo, so a hijacked action can silently run malicious code with our workflow tokens — the `tj-actions/changed-files` compromise worked exactly this way. SHAs freeze what runs.
- Dependabot's existing `github-actions` config bumps pinned SHAs the same way it bumps tags, so the pins won't go stale.
- Resolves all 10 open CodeQL `actions/unpinned-tag` alerts from the security-extended suite.

## Test plan
- [x] `rg 'uses:.*@v\d'` finds no remaining tag-pinned actions
- [x] Each SHA resolved directly from the action repo's tag via the GitHub API
- [ ] CI on this PR exercises checkout/setup-go/setup-node/msys2 pins; release-only pins get exercised on the next release

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -28,17 +28,17 @@ jobs:
               GOARCH: arm64
 
     steps:
-    - uses: actions/checkout@v7
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
 
     - name: Set up Go
-      uses: actions/setup-go@v7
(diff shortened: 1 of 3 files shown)


CtrlSpice:
Decided against SHA pinning — readability cost isn't worth it for this project.

PR #381, opened 2026-08-22
Stop telling readers a finished trace is still arriving

CtrlSpice:
## What

- `No root span yet` → `No root span` in the trace list card

## Why

- `hasRootSpan` means only "some span has a null parent". Two ordinary shapes have none and are not waiting: a trace whose root is absent (spans promoted to roots, renders fine) and a trace whose parent links form a cycle (recovered by the salvage walk in #379, rendered with cycle marks)
- "yet" told the reader to wait for something that will never arrive

Frontend suite green (515).


CtrlSpice:
Superseded by #383, which carries this change — the branch conflicted with main only in built assets, which is not worth a rebase for a one-line label.

PR #384, opened 2026-08-22
Stop installing a Windows toolchain nothing compiles with

### LINKED ISSUE #324: P3: Windows CI spends ~74s extracting MSYS2 before it builds
After Go build caching and `update: false`, the Windows job is ~176s against
macOS's 40s. Build is down to 20s and tests to 35s; the remainder is MSYS2 base
extraction, which the package cache does not cover.

**Proposed:** try caching the extracted toolchain directory itself, or compare
against a prebuilt container.

CtrlSpice:
Fixes #324.

## The issue's premise was wrong

It attributes the Windows job's cost to MSYS2 base extraction. The step timings put it elsewhere:

| phase | time |
|---|---|
| download (already in toolcache) | 0.8s |
| extract base | 8.0s |
| package cache restore — a hit | 2.4s |
| **pacman installs 43 packages** | **48.4s** |

`cache: true` was already on and the key already hit. The cost is installing `mingw-w64-ucrt-x86_64-toolchain`, a meta-package, on every run.

## And the toolchain was never used

Probed on a real run rather than reasoned about:

- **Before** setup-msys2: neither `C:\msys64\ucrt64\bin\gcc.exe` nor `RUNNER_TEMP\msys64\ucrt64\bin\gcc.exe` exists
- **After** setup-msys2 and the `C:\msys64\ucrt64\bin` PATH line: `gcc resolves to C:\mingw64\bin\gcc.exe`

cgo has been using the runner image's preinstalled mingw64 the whole time. The 61s produced a compiler nothing invoked.

## What this does

- Removes the setup-msys2 step and the PATH line
- Replaces them with a step that checks gcc exists and fails naming the fix

The trade is a dependency on the runner image. That is what the check is for: without it, a removed preinstall surfaces as a cgo error deep in a DuckDB build, which reads like a code problem rather than an image change.

## Verified

Windows job green with no MSYS2 at all — build and the full `go test ./...`, which is where DuckDB's cgo actually gets exercised. The MSYS2 step goes 61s → 0. Wall-clock was 164s → 145s on these two runs, but `Set up Go` varied 41s → 74s between them, so treat the step number as the real one and the total as noisy.

## Not touched

`release.yml` and `test-go-install.yml` carry the same pattern, and `release.yml` uses `update: true`, so it pays even more. Left alone deliberately — that job builds the binaries people download, and switching its toolchain is a separate decision from CI.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -68,6 +68,24 @@ jobs:
     # How: flip this to true for one run and put it back, or delete the
     # msys2-pkgs-* entry from the Actions cache. Editing the install list below
     # also re-keys the cache on its own, since the key hashes that list.
+    # TEMPORARY (#324): which gcc does the build actually use?
+    # The PATH step below names C:\msys64, the runner's preinstalled MSYS2,
+    # while setup-msys2 installs under RUNNER_TEMP. If the preinstalled one
+    # already carries a ucrt64 gcc, the setup step's 61s buys nothing.
+    - name: Probe preinstalled toolchain (Windows)
+      if: matrix.os == 'windows-latest'
+      run: |
+        Write-Host "RUNNER_TEMP = $env:RUNNER_TEMP"
+        foreach ($root in @("C:\msys64", "$env:RUNNER_TEMP\msys64")) {
+          $gcc = Join-Path $root "ucrt64\bin\gcc.exe"
+          if (Test-Path $gcc) {
+            Write-Host "FOUND $gcc"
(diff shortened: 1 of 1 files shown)


CtrlSpice:
Superseded by #385. The change here was right — setup-msys2 was 61s of waste — but the reason was incomplete: I read C:\mingw64 as an accident to be tolerated. It is actually the only compiler on the image that can link duckdb-go-bindings' prebuilt libduckdb_static.a, so #385 pins it deliberately and applies the same steps to release and go-install so they cannot drift.

PR #415, opened 2026-08-27
Key a span by its trace and its id

CtrlSpice:
Stacked on #414 — merge that first and this retargets to `main` on its own. **Schema 7 → 8.**

## Fixed

- A span id is only required to be unique *within its trace*. Keying `spans` on `span_id` alone was stricter than OTLP and rejected conformant senders — fixtures numbering spans per trace, replay tools, any deterministic generator. Their second trace looked like a duplicate of the first. `spans` is now keyed on `(trace_id, span_id)`.
- `events` and `links` carry the owning trace and foreign-key the pair. Every join from a child to its span carries the trace too: without it, two traces sharing a span id contaminate each other.
- `links.trace_id` → `linked_trace_id`. It always meant the context pointed *at*, and now reads correctly beside `linked_span_id`.

## What the test caught

Two traces each using span ids `1` and `2`. With the key changed but the queries not, fetching one trace returned **all four spans**, both traces' events, and `unplacedSpanCount: -2`. The offenders were the tree walk's payload join (`join spans s on s.span_id = st.span_id`) and five child predicates. The predicates now share `eventOwner` / `linkOwner` constants so a sixth cannot forget.

## Notes

- Wire format unchanged — `link_json` still emits `traceID` for the linked context. Frontend suite passes untouched (587 tests, no edits).
- Anyone with an existing `--db` file gets `ErrSchemaIncompatible` on open. Documented behaviour for a version bump, but worth knowing it lands right after a release.
- `deleteSpansByIDs` is now ambiguous on the wire: a span id no longer identifies one span, so it deletes across traces. Existing behaviour preserved rather than silently changed; taking pairs is a follow-up.
- SQL goldens regenerated; the diff is only trace scoping.

### DIFF
--- a/desktopexporter/internal/store/queries/ddl/indexes/idx_events_span.sql
+++ b/desktopexporter/internal/store/queries/ddl/indexes/idx_events_span.sql
@@ -1,1 +1,1 @@
-create index if not exists idx_events_span on events(span_id)
+create index if not exists idx_events_span on events(trace_id, span_id)
--- a/desktopexporter/internal/store/queries/ddl/indexes/idx_links_span.sql
+++ b/desktopexporter/internal/store/queries/ddl/indexes/idx_links_span.sql
@@ -1,1 +1,1 @@
-create index if not exists idx_links_span on links(span_id)
+create index if not exists idx_links_span on links(trace_id, span_id)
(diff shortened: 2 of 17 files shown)


CtrlSpice:
Superseded by #416 — this closed automatically when `fix/lossy-appender-ingest` was deleted on merging #414. Same branch, rebased onto main.

PR #422, opened 2026-08-28
feat: page through status errors in the waterfall

### LINKED ISSUE #411: [feature] Allow paging through the errors
### What problem does this solve?

I have a 2000+ span trace with 7 errors. But I have to scroll to find them. 

### What would make this better for you?

It would be nice if I could click a button to jump to the next / prev one

### Where would this live?

- [x] Traces
- [ ] Metrics
- [ ] Logs
- [ ] Cross-signal
- [ ] Not sure / general

### Anything else?

_No response_

### Tell me a dad joke

_No response_

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](../blob/main/CODE_OF_CONDUCT.md).

mvanhorn:
## Summary

The waterfall header already shows an error count, but there was no way to move between those spans, so locating each error in a large trace meant scrolling the whole list.

This adds previous/next controls beside the error badge that move the selection to the nearest error span in waterfall order.

## Why this matters

Closes #411

`headerErrorCount` in `WaterfallView.svelte` was computed by filtering `spans` for `statusCode === 'Error'` and then discarding the matches, keeping only the length. The information needed to navigate was already being derived and thrown away; this keeps the filtered list and drives the controls from it.

Four behaviours are deliberate:

- The controls disable at each end instead of wrapping, so paging cannot silently loop back to the first error.
- Paging from a healthy span jumps to the nearest error in that direction rather than restarting from the top.
- Only spans whose `statusCode` is `Error` count, so navigation agrees with the badge. Exception-only spans are excluded, which is covered by a test.
- Selection goes through the existing `onSelectSpan` scroll path, so a branch the reader collapsed is not reopened.

## Testing

`npx vitest run src/components/traces/Waterfall/WaterfallView.test.ts` -> 14 passed.

The added cases pin the behaviour rather than just exercising it. Reverting only `WaterfallView.svelte` to `upstream/main` and keeping the tests:

```
❯ src/components/traces/Waterfall/WaterfallView.test.ts (14 tests | 5 failed)
  × walks status errors forward and backward in waterfall order
  × chooses the nearest directional status error from a healthy span
  × disables unavailable directions without wrapping or selecting
  × excludes exception-only spans so navigation agrees with the error badge
  × uses selection scrolling without reopening the reader's collapsed branch
Tests  5 failed | 9 passed (14)
```

Not tested: no manual pass in the running desktop UI, so the control placement and icon sizing in the header have not been reviewed visually. The assertions cover navigation logic, not appearance.

AI was used for assistance.


CtrlSpice:
Thank you for this! Was seriously a lovely thing to find in the queue. Cheers!

PR #528, opened 2026-09-19
Preserve complete resource payload identity

CtrlSpice:
- Stacked on the exact reviewed head of #527 (`8148214`), which is now merged via a squash commit; retaining its branch as the base preserves the approved history because a direct merge of this exact candidate into current `main` conflicts at the schema-version history.
- Derives stored resource identity from the complete canonical typed attribute payload plus `droppedAttributesCount`, preserving exact span, log, and metric-ingest resource references while deduplicating identical payloads across batches and signals.
- Derives metric series identity from originating resource attributes and datapoint attributes, so resource attribute changes split series while dropped-count-only changes retain distinct resource rows without fragmenting the series.
- Advances the no-migration schema contract to v15; previously lost resource associations cannot be reconstructed from older captures.
- Projects only identifying resource attributes at series level with a synthetic zero dropped count, while the top-level metric resource remains the complete exact payload from the selected representative ingest.
- Independent exact-head review approved base `8148214` through head `80adbb4` with no findings (patch SHA-256 `b8e06b2fd414810d6d2f07a5c6f8ee708bf9a0cc563cf6c225e3ed2ba49a0c45`).


CtrlSpice:
Superseded by #529. PR #527 merged while #528 was being published, so the byte-identical resource payload repair was replayed from merged main and independently re-reviewed at the replacement head.

